PR #85, opened 2015-11-04
Emfatten header; slim it when we're not in the home view

akx:
Refs #3


akx:
Oops, I thought this had been merged. Anyway, rebasing~

PR #86, opened 2015-11-04
Add Hearing view mockup

rphlo:


akx:
The end result looks great code-wise! :lollipop: 

Though: Rebase to resolve conflicts and while you're at it , squash e5cc265 with d5faf85? I mean, no reason to have one commit adding Leaflet in bower and another removing it to add it via npm :)

akx:
Yay!

PR #88, opened 2015-11-04
Improve Hearing page

rphlo:


src/components/HearingList.js
```diff
@@ -5,7 +5,7 @@ import {injectIntl} from 'react-intl';
 class HearingListItem extends React.Component {
   render() {
     const hearing = this.props.hearing;
-    return <li>{hearing.id}</li>;
+    return <li><a href={"/hearing_"+hearing.id}>{hearing.heading}</a></li>;
```
akx: This should use a [react-router `<Link>`](https://github.com/rackt/react-router#whats-it-look-like) component.

src/views/Hearing/index.js
```diff
@@ -20,9 +20,9 @@ class Hearing extends React.Component {
   }
 
   getLabels() {
-    return (<div>
-      <Label>label 1</Label> <Label>label 2</Label>
-    </div>);
+    const {hearingId} = this.props.params;
```
akx: Actually, maybe just pass `data` to this function from `render()`? :)

src/views/Hearing/index.js
```diff
@@ -45,15 +45,21 @@ class Hearing extends React.Component {
   }
 
   getOverviewMap() {
-    const position = [51.505, -0.09];
+    const {hearingId} = this.props.params;
```
akx: Same here: just pass `data` to this function from `render()`? :)

akx:
Very nice! :)

PR #90, opened 2015-11-05
Use moment library for time display

rphlo:


src/views/Hearing/index.js
```diff
@@ -53,7 +54,7 @@ class Hearing extends React.Component {
       <Col xs={6} sm={3}>
         <div>
           <h4><FormattedMessage id="timetable"/></h4>
-          <i className="fa fa-clock-o"></i> <FormattedMessage id="closing"/> {data.close_at}
+          <i className="fa fa-clock-o"></i> <FormattedMessage id="closing"/> {moment(data.close_at).fromNow()}
```
akx: Hmm... [`react-intl` already provides `<FormattedRelative/>` for relative date formatting](https://github.com/yahoo/react-intl/blob/master/src/components/relative.js), would that work?

PR #91, opened 2015-11-06
Add Hearing Images Components

rphlo:


src/components/HearingImage.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+class HearingImage extends React.Component {
+  render() {
+    const {data} = this.props;
+    return (<div>
+        <img width="100%" title={data.title} alt={data.title} src={data.url}/>
```
akx: [`className="img-responsive"`](http://getbootstrap.com/css/#images-responsive) maybe?

src/components/HearingImageList.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import HearingImage from './HearingImage';
+
+class HearingImageList extends React.Component {
+  render() {
+    const {images} = this.props;
+    return (<div>
+      {images.map((image) => <HearingImage data={image}/>)}
```
akx: [Arrays of components must have `key` props.](http://facebook.github.io/react/docs/multiple-components.html#dynamic-children)

src/components/HearingImageList.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import HearingImage from './HearingImage';
+
+class HearingImageList extends React.Component {
+  render() {
+    const {images} = this.props;
+    return (<div>
+      {images.map((image, index) => <HearingImage data={image} key={index}/>)}
```
akx: Maybe use the image URL as the key?

I mean, I doubt using an index will do the Right Thing during DOM reconciliation, as if there were, say, images "0", "1" and "2", and you'd delete the one formerly called "1", you'd then get "0" and "1", and React would think "2" has gone away.

PR #97, opened 2015-11-10
Add css to match prototype look and enable comments

rphlo:


src/components/CommentList.js
```diff
@@ -7,13 +7,13 @@ class CommentList extends React.Component {
...
     if (comments.length === 0) {
-      return (<div>
+      return (<div id="hearing-comments">
         <h2><FormattedMessage id="comments"/></h2>
         {areCommentsOpen ? <CommentForm/> : null}
         <p><FormattedMessage id="noCommentsAvailable"/></p>
       </div>);
     }
-    return (<div>
+    return (<div id="hearing-comments">
```
akx: May there ever be more than one `CommentList` component on a page? If so, can't (legally) use `id`

PR #102, opened 2015-11-13
View scenario comments

rphlo:


akx:
Can you rebase or something to get rid of the master merge, please? :)

src/components/Scenario.js
```diff
@@ -13,9 +13,18 @@ export default class Scenario extends React.Component {
   }
 
   toggle() {
+    if (this.state.collapsed === true) {
```
akx: I don't think the toggle action is the right place to (re)load comments. Maybe the [componentWillUpdate](http://facebook.github.io/react/docs/component-specs.html#updating-componentwillupdate) method? You could check if `!nextState.collapsed` and load comments then.

PR #105, opened 2015-11-16
Add initial test setup

akx:
Took several hours of head-scratching before I realized I'd accidentally installed isparta@4.0.0, which only works with Babel 6 and fails in mysterious ways with Babel 5.x.


akx:
Tests fail because of a missing envvar, but I'm gonna see if I can work around that from inside Node.

PR #110, opened 2015-11-19
Fix voting for scenario comments

rphlo:


akx:
Would be nice to have a test for this, though to be absolutely honest I don't have any off-the-cuff idea on how to actually implement one :)

PR #137, opened 2016-02-15
Tune closure info

tuomas777:
Added Classnames v2.2.3

Added closure info section handling. In case closure info section exists, it is rendered as the topmost
section with yellow background.

### DIFF
--- a/src/components/Section.js
+++ b/src/components/Section.js
@@ -1,6 +1,8 @@
 import React from 'react';
 import CommentList from './CommentList';
 import Icon from 'utils/Icon';
+import {isSpecialSectionType} from 'utils/section';
+import classNames from 'classnames';
 
 export default class Section extends React.Component {
   constructor(props) {
@@ -39,6 +41,13 @@ export default class Section extends React.Component {
     if (section.type === "introduction") { // Intros never render this
       return null;
     }
+    if (section.type === "closure info") {
+      return (
+        <h3 className="section-title">
+          {section.title}
+        </h3>
+      );
+    }
     const iconName = (collapsed ? "chevron-right" : "chevron-down");
     return (
       <h3 className="section-title" onClick={this.toggle.bind(this)}>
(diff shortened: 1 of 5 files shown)


src/components/Section.js
```diff
@@ -49,7 +57,7 @@ export default class Section extends React.Component {
 
   render() {
     const {section} = this.props;
-    const collapsed = this.state.collapsed && (section.type !== "introduction");
+    const collapsed = this.state.collapsed && !["introduction", "closure info"].includes(section.type);
```
akx: The `!["introduction", "closure info"].includes(section.type)` fragment is now duplicated twice... maybe move it into a helper fn?

src/views/Hearing/index.js
```diff
@@ -120,7 +120,8 @@ class Hearing extends React.Component {
     const hearingAllowsComments = !hearing.closed && (new Date() < new Date(hearing.close_at));
     const onPostVote = this.onVoteComment.bind(this);
     const introSection = detect(hearing.sections, (section) => section.type === "introduction");
-    const regularSections = hearing.sections.filter((section) => section.type !== "introduction");
+    const closureInfoSection = detect(hearing.sections, (section) => section.type === "closure info");
+    const regularSections = hearing.sections.filter((section) => !["introduction", "closure info"].includes(section.type));
```
akx: This could now be refactored to use [_.difference](https://lodash.com/docs#difference):

```js
const regularSections = difference(hearing.sections, [introSection, closureInfoSection]);
```

:)

src/utils/section.js
```diff
@@ -0,0 +1,5 @@
+const specialSectionTypes = ["introduction", "closure info"];
+
+export default function isSpecialSectionType(sectionType) {
```
akx: This should not be the default export... A regular export only.

tuomas777:
Changed the export to a regular one.

akx:
fangz :D

PR #143, opened 2016-02-24
Dynamic section type

tuomas777:
The sections are grouped by type. Sidebar contains a link to each
group. The groups are ordered based on the ordering of their most
lowest ordered section.

Added also missing translation of "no comments available".

### DIFF
--- a/src/views/Hearing/Sidebar.js
+++ b/src/views/Hearing/Sidebar.js
@@ -11,24 +11,13 @@ import formatRelativeTime from '../../utils/formatRelativeTime';
 import Icon from 'utils/Icon';
 
 class Sidebar extends React.Component {
-  getScenariosCount(sections) {
-    const scenarios = sections.filter(function(section) {
-      return section.type === "scenario";
-    });
-    return scenarios.length;
-  }
-
   render() {
     const {hearing} = this.props;
+    const {sectionGroups} = this.props;
     const boroughDiv = (hearing.borough ? (<div>
       <h4><FormattedMessage id="borough"/></h4>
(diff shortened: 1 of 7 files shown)


src/views/Hearing/Sidebar.js
```diff
@@ -11,24 +11,13 @@ import formatRelativeTime from '../../utils/formatRelativeTime';
...
-  getScenariosCount(sections) {
-    const scenarios = sections.filter(function(section) {
-      return section.type === "scenario";
-    });
-    return scenarios.length;
-  }
-
   render() {
     const {hearing} = this.props;
+    const {sectionGroups} = this.props;
```
akx: Consolidate these to `const {hearing, sectionGroups} = this.props;`.

src/views/Hearing/index.js
```diff
@@ -121,11 +121,28 @@ class Hearing extends React.Component {
     const hearingAllowsComments = !hearing.closed && (new Date() < new Date(hearing.close_at));
     const onPostVote = this.onVoteComment.bind(this);
     const introSection = detect(hearing.sections, (section) => section.type === "introduction");
-    const closureInfoSection = detect(hearing.sections, (section) => section.type === "closure info");
+    const closureInfoSection = detect(hearing.sections, (section) => section.type === "closure-info");
     const regularSections = hearing.sections.filter((section) => !isSpecialSectionType(section.type));
+
+    // group sections by type
+    const sectionGroups = [];
```
akx: This logic could probably be refactored into a function of its own.

PR #144, opened 2016-02-24
Mapdon plugin (and plumbing)

### LINKED ISSUE #131: Implement possibility for 3rd party add-ons to a hearing, as e.g. sections in a hearing
Pilot case code: http://dev.mapdon.com/hkr/map.php

akx:
Fixes #131


akx:
So the tests don't currently work -- that'll have to be fixed at some point.

PR #163, opened 2016-03-14
Add feedback prompt

Rikuoja:


src/views/Home.js
```diff
@@ -45,6 +45,13 @@ class Home extends React.Component {
             <HearingList hearings={hearingLists.newestHearings} />
           </div>
         </Col>
+        <Col md={4} xs={12}>
+          <div className="feedback-box">
+            <a href="mailto:dev@hel.fi">
```
akx: Would it maybe be a good idea to add a [`?subject=` parameter](http://stackoverflow.com/a/4782084/51685)? Something like "Kerro Kantasi -palaute"?

akx:
VERY FEED
SUCH BACK
WOW

PR #164, opened 2016-03-15
Post comment authorization code from url parameter

Rikuoja:


akx:
Seems straightforward enough (though I'd still like that spec/UC for this whole auth code thing to be referenced with these changes :) )

Issue #139, opened 2016-02-18
Instructions and button for adding a comment are missing.


Rikuoja:

Rikuoja:
At least for closed hearing. Add explanatory text?

Issue #167, opened 2016-03-18
Comments on sections are not visible. Therefore, many imported comments on old hearings are no longer displayed.


Rikuoja:

Rikuoja:
Serialize the section comments when serializing the hearing, or fetch them separately when loading a section?

PR #177, opened 2016-03-30
Display section thumbnails and info for closed hearings

Rikuoja:


Rikuoja:
Closes #139 #153 #167

Rikuoja:
And #178

src/views/Hearing/index.js
```diff
@@ -128,12 +129,20 @@ class Hearing extends React.Component {
     const hearingAllowsComments = !hearing.closed && (new Date() < new Date(hearing.close_at));
     const onPostVote = this.onVoteComment.bind(this);
     const introSection = find(hearing.sections, (section) => section.type === "introduction");
-    const closureInfoSection = find(hearing.sections, (section) => section.type === "closure-info");
+    let closureInfoSection = find(hearing.sections, (section) => section.type === "closure-info");
```
akx: Would probably be better to let this remain `const` and move the rvalue and the fallback logic (below) into a helper function like `_getClosureInfoSection()`.

akx:
Also, hmm, are we using a `develop` branch instead of feature branches? :open_mouth:

Rikuoja:
Fair point, I'll do features in separate branches in the future.

src/views/Hearing/index.js
```diff
@@ -128,12 +145,12 @@ class Hearing extends React.Component {
     const hearingAllowsComments = !hearing.closed && (new Date() < new Date(hearing.close_at));
     const onPostVote = this.onVoteComment.bind(this);
     const introSection = find(hearing.sections, (section) => section.type === "introduction");
-    const closureInfoSection = find(hearing.sections, (section) => section.type === "closure-info");
+    const closureInfoSection = this.getClosureInfo();
```
akx: Maybe just pass `hearing` in as a param instead of re-deriving it in the helper? :)

akx: (As is done with `getOpenGraphMetaData` etc already)

PR #181, opened 2016-04-04
Add report link to hearing

Rikuoja:


src/views/Hearing/index.js
```diff
@@ -188,6 +190,8 @@ class Hearing extends React.Component {
               onPostVote={onPostVote}
             />
           </div>
+          <hr/>
+          <a href={reportUrl}><FormattedMessage id="downloadReport"/></a>
```
akx: This is visible to everyone? Is that actually okay?

Rikuoja: Yes, the data is open and available on the page and in the api, so there is no reason the report should not be downloadable too.

PR #182, opened 2016-04-05
Add piwik code to helmet

Rikuoja:


akx:
Seems okay.

akx:
How about those feature branches btw? ;)

PR #183, opened 2016-04-06
Track all routes in Piwik

Rikuoja:


src/routes.js
```diff
@@ -8,13 +8,25 @@ import Home from 'views/Home';
 import Info from 'views/Info';
 import NotFound from 'views/NotFound';
 
+function trackLink() {
+  /* global Piwik */
+  try {
+    const tracker = Piwik.getTracker("http://analytics.hel.ninja/piwik/piwik.php", 2);
```
akx: The tracking URL should be configurable. After all, this is supposed to be a reusable app...

Rikuoja: K, what would be the best location for the tracking URL? All the way in kk.config and then get it to window via props?

akx: Yeah. Maybe an `uiConfig` object in `kk.config` that gets plonked into the HTML, and is then read by the frontend code?

akx:
So yeah, `getOptions` in `getSettings()` should check that if `_.isString(uiConfig)`, it's `JSON.parse(uiConfig)`'d at that point. That'll retain compatibility with envvar-based configuration.

src/config.js
```diff
@@ -1,5 +1,6 @@
 const config = {
   languages: ['fi'],
-  apiBaseUrl: (typeof window !== 'undefined' ? window.API_BASE_URL : null) || 'http://localhost:8000/'
+  apiBaseUrl: (typeof window !== 'undefined' ? window.API_BASE_URL : null) || 'http://localhost:8000/',
+  uiConfig: (typeof window !== 'undefined' ? window.UI_CONFIG : null) || null
```
akx: This does not need the ` || null` fallback. Also, add the trailing comma.

src/config.js
```diff
@@ -1,6 +1,6 @@
 const config = {
   languages: ['fi'],
   apiBaseUrl: (typeof window !== 'undefined' ? window.API_BASE_URL : null) || 'http://localhost:8000/',
-  uiConfig: (typeof window !== 'undefined' ? window.UI_CONFIG : null) || null
+  uiConfig: typeof window !== 'undefined' ? window.UI_CONFIG : null,
```
Rikuoja: My IDE is vehemently against unneeded commas.

PR #198, opened 2016-04-11
Sensibly show total comment numbers and section comment numbers

Rikuoja:


assets/less/hel/hearing-page.less
```diff
@@ -37,6 +37,11 @@
     .list-group .badge {
         font-size: 1.1rem;
     }
+
+    .commenticon {
```
akx: `comment-icon`? Unless this is some sort of a decepticon.

src/views/Hearing/Sidebar.js
```diff
@@ -29,6 +29,15 @@ class Sidebar extends React.Component {
         <div className="hearing-sidebar">
           <Row>
             <Col sm={6} md={12}>
+              <div className="sidebar-section commentNumber">
+                <Icon name="comment-o"/> {' '}
+                <FormattedMessage id="aTotalOf"/> {' '}
```
akx: Maybe this "Total X comments" thing should be a single message... Something like
```
<FormattedPlural value={hearing.n_comments}
    one={<FormattedMessage id="totalSubmittedComment" values={{n: hearing.n_comments}}/>}
    other={<FormattedMessage id="totalSubmittedComments" values={{n: hearing.n_comments}}/>}
/>
```
and
```
{
  "totalSubmittedComment": "Yhteensä {n} mielipide",
  "totalSubmittedComments": "Yhteensä {n} mielipidettä",
}
```

or however the exact syntax goes? :)

src/views/Hearing/index.js
```diff
@@ -225,12 +226,14 @@ function groupSections(sections) {
     const sectionGroup = find(sectionGroups, group => section.type === group.type);
     if (sectionGroup) {
       sectionGroup.sections.push(section);
+      sectionGroup.n_comments += section.n_comments;
```
akx: You may not mutate the data passed in!

Rikuoja: This is not mutating passed in data. The section group is created in the very same function, by iterating through the sections that have been passed.

akx: It's still super duper unclean for a child component to mutate a prop.

Could this calculation be done in the `groupSections()` fn instead?

Rikuoja: As said, the child component does not mutate a prop. This is constructing a new prop for a child component, and it is done within a single function in a single component.

Rikuoja: And, as you suggested, the calculation is done inside the groupSections() fn.

akx: Right. I must've been spectacularly out of it while trying to review this. Is fine.

PR #199, opened 2016-04-13
Show disclaimer when posting comments

Rikuoja:


akx:
We're already shipping React-Markdown or similar, aren't we? Maybe just make the entire string a translatable fragment of Markdown?

Rikuoja:
Hm, couldn't find React-Markdown in the app?

akx:
Ah. Right, we're [using markdown-loader with Webpack](https://github.com/City-of-Helsinki/kerrokantasi-ui/blob/master/src/views/Info/index.js) . . . should switch that to react-markdown instead :)

PR #202, opened 2016-05-02
Add build script for npm

vikoivun:
Idea here is to have a single preparation command "npm run build". This command is supposed to performs all steps beyond "npm i", that are necessary to prepare the codebase for running within a process manager.
npm also supports lifecycle events for automating this, but I cannot tell if using them would cause difficulties elsewhere.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "main": "index.js",
   "scripts": {
     "start": "node server --server-rendering",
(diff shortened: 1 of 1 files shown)


akx:
Looks good to me. (`npm run build` just so happens to be the same convention we use at Anders.)

Issue #204, opened 2016-05-04
Prevent accidental scrolling and zooming of hearing map outside the hearing area


Rikuoja:

JsseL:
So #207 would fix this. After trying it in action, I'm not sure if it actually makes the map more user friendly. Leaflet's max bounds implementation feel a bit glitchy and it also felt a bit too restricting.

I tried to make it a bit more user friendly by setting the max bounds a bit wider than the hearing area itself. This way user is able to pan the viewport a bit wider than just the border of the hearing area. Increasing the max bounds even more could make it even more usable, but eventually it would lead to the original problem.

I think you should try it out and see how it feels in action.

JsseL:
Updated the PR. Now it should actually improve the UX of the map element.

PR #207, opened 2016-05-10
Set max bounds to hearing map

JsseL:
Properly set max bounds makes sure that the hearing map can't be
accidentally scrolled or zoomed out of the Helsinki area.

This should improve the user experience of the map component.

Refs: #204

### DIFF
--- a/src/components/OverviewMap.js
+++ b/src/components/OverviewMap.js
@@ -47,7 +47,14 @@ class OverviewMap extends React.Component {
           ref={(input) => {
             if (!input) return;
             const bounds = input.getLeafletElement().getBounds();
-            if (bounds.isValid()) input.props.map.fitBounds(bounds);
+            if (bounds.isValid()) {
+              input.props.map.fitBounds(bounds);
+              const viewportBounds = [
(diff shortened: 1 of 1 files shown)


JsseL:
I modified the restricting bounding to be the roughly the Helsinki area instead of the specific hearing area. This change made the map actually pretty nice to use.

In addition to the bounding box, I added min zoom level (see the second commit).

JsseL:
### Note
Defined bounds for Helsinki are just rough estimates which felt ok. If someone can provide more accurate coordinates, feel free to inform me and I'll add them.

Rikuoja:
```
                [60.1, 24.75],  // SouthWest corner
                [60.35, 25.3]  // NorthEast corner
```

would seem like suitable bounds to me.

Rikuoja:
Otherwise LGTM :)

JsseL:
Changed bounds to the ones you suggested. They indeed felt good!

JsseL:
I think we got the coordinates a bit wrong after all. The bounds are too restrictive at the moment. So this needs small fine tuning of the bounds.

Issue #174, opened 2016-03-22
When adding a comment, reload comments and scroll to display the latest comment


Rikuoja:

Rikuoja:
Currently the comment is not shown when posted, and the written text remains in the textbox. This results in people posting the same comment multiple times, expecting to get visual feedback.

JsseL:
Partial fix for the problem. Will implement the scroll to latest comment feature later.

Rikuoja:
If the latest comment is displayed first, no scrolling is needed. Currently, the comment is appended to the end of the list so it is not displayed?

JsseL:
Yep, currently shown in the bottom.

PR #211, opened 2016-05-23
Clear message text field after submit

JsseL:
According to issue #174 preserved text in comment box is causing
bad usability.

After comment submit remove the text

Refs: #174

### DIFF
--- a/src/components/BaseCommentForm.js
+++ b/src/components/BaseCommentForm.js
@@ -6,9 +6,39 @@ import Icon from 'utils/Icon';
 import CommentDisclaimer from './CommentDisclaimer';
 
 class BaseCommentForm extends React.Component {
-  constructor(props) {
-    super(props);
+  constructor(props, context) {
+    super(props, context);
     this.state = {collapsed: true, commentText: ""};
+  }
+
+  componentDidMount() {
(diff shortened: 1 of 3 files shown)


akx:
If the submission fails, the user will have lost their comment. :(

Rikuoja:
Granted, the clear should be done only when submit succeeds and we are reloading.

JsseL:
This required some helping hand from the dark master.

akx:
I prefer the epithet "dork master".

akx:
@JsseL Ah, we need to turn

```javascript
(window.devToolsExtension ? window.devToolsExtension() : identity)
```
into
```javascript
(typeof window !== 'undefined' && window.devToolsExtension ? window.devToolsExtension() : identity)
```

JsseL:
Pair coded the solution with @akx so to us this is ready to be shipped. Project doesn't have any UI functionality testing so coverage decrease can either be ignored or we can start doing some functionality testing too, if it is seen worthwhile.

@riku your call.

JsseL:
Some reason why we wouldn't merge this already?

Rikuoja:
Looking good!

PR #218, opened 2016-06-07
Changes needed because of backward incompatible API changes

tuomas777:
- introduction section is renamed to main

- hearing images are converted to main section images, hearings have
  also main_image field within their data

- hearing comments are converted to main section comments

ATTENTION requires the corresponding changes to be deployed on the backend!!

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -70,20 +70,6 @@ export function fetchSectionComments(hearingId, sectionId) {
   };
 }
 
-export function postHearingComment(hearingId, text, authCode = "") {
-  return (dispatch, getState) => {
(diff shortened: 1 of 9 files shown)


akx:
Tbh these should all be separate commits and preferably they should have a "Refs" to the corresponding backend issue/PR/commit :)

akx:
If you feel like separating the changes to atomic commits, then please do.

The code lgtm though.

tuomas777:
I have a feeling that won't be going as in Strömsö, but I'll see what I can do :)

tuomas777:
Updated, divided the commit to four commits

akx:
Thanks 👍

PR #219, opened 2016-06-08
Allow sections be not commentable when the hearing isn't

### LINKED ISSUE #216: Let single sections not be commentable even if the hearing is commentable
Currently, hearingAllowsComments for the whole hearing is passed directly onto SectionList and all Sections.

Therefore, all sections look like commentable on the UI even if they are not commentable in the API. Trying to post a Comment on such a section results in an API error.

tuomas777:
Closes #216

### DIFF
--- a/src/components/Section.js
+++ b/src/components/Section.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import CommentList from './CommentList';
 import Icon from 'utils/Icon';
-import {isSpecialSectionType} from 'utils/section';
+import {isSpecialSectionType, userCanComment} from 'utils/section';
 import classNames from 'classnames';
 import MapdonHKRPlugin from './plugins/mapdon-hkr';
 import Alert from 'react-bootstrap/lib/Alert';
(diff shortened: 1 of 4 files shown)


src/components/SectionList.js
```diff
@@ -41,6 +42,7 @@ SectionList.propTypes = {
   sections: React.PropTypes.array,
   sectionComments: React.PropTypes.object,
   nComments: React.PropTypes.number,
+  user: React.PropTypes.object
```
akx: Add a comma.

tuomas777:
added a comma

src/components/Section.js
```diff
@@ -168,4 +168,5 @@ Section.propTypes = {
   onPostVote: React.PropTypes.func,
   loadSectionComments: React.PropTypes.func,
   comments: React.PropTypes.object,
+  user: React.PropTypes.object
```
akx: This could use that trailing comma too. :D

tuomas777:
added a second comma

PR #222, opened 2016-06-21
Allow non authenticated users set comment author nickname

### LINKED ISSUE #208: Allow anonymous users to give nickname when commenting

tuomas777:
Closes #208


src/components/BaseCommentForm.js
```diff
@@ -69,11 +75,22 @@ class BaseCommentForm extends React.Component {
...
     if (!this.state.collapsed) {
       return (<div className="comment-form">
         <form>
           <h3><FormattedMessage id="writeComment"/></h3>
           <Input type="textarea" value={this.state.commentText} onChange={this.handleTextChange.bind(this)}/>
+          {canSetNickname ? <h3><FormattedMessage id="nickname"/></h3> : null}
+          {canSetNickname ? (
+            <Input
+              type="text"
+              placeholder={this.props.intl.formatMessage({id: "anonymous"})}
```
akx: Missing translation for "anonymous"?

akx:
So uhm, are authenticated users shown somehow differently in the UI? That is, if I post as an authenticated "Aarni Koskela", and then some joker goes and uses "Aarni Koskela" as their nickname to post something nasty, will the audience be able to see that these are probably not the same people?

Rikuoja:
An excellent point right there. Obviously, authenticated users will need to be differentiated in the UI. Any suggestions on what would be the user-friendly way to show the difference?

src/components/BaseCommentForm.js
tuomas777: https://github.com/City-of-Helsinki/kerrokantasi-ui/blob/master/src/i18n/fi.json#L41

tuomas777:
Yeah that would be good. It needs some work on the backend too, because I don't think this information is currently present there either

JsseL:
Maybe some kind of user icon next to the nickname for logged in users?

Rikuoja:
Yeah, though the icon should be kinda self-evident, otherwise people will think it's the same user nevertheless. Perhaps an added "registered user" tag or something?

tuomas777:
Also if authenticated users cannot change their nicknames, probably non authenticated users should not be able to use those exact nicknames?

akx:
Unassigned myself pending the changes in the comments.

tuomas777:
Updated, added registered user differentiating when showing comments.

tuomas777:
kommenttiboksi

tuomas777:
kommentit

PR #232, opened 2016-07-04
Looser map bounds

JsseL:
Finetune map views with better params.

Refs: #204


JsseL:
@Rikuoja Currently there are slight problems with the bounds. See for example this [page](https://kerrokantasi.hel.fi/map). These small changes will make life of Kerrokantasi users better.

Issue #227, opened 2016-06-28
As a developer, I want to add a plugin with option to turn commenting off, for visualizing existing comments.


Rikuoja:

Rikuoja:
Needed for KSV citybikes questionnaire.

PR #233, opened 2016-07-07
The upcoming KSV Citybikes questionnaire.

Rikuoja:
Contains all the needed functionality for implementing Citybikes questionnaire and #227.


src/components/plugins/mapdon-ksv.js
```diff
@@ -0,0 +1,99 @@
...
+        comments,
+        instanceId: self.pluginInstanceId
+      });
+    }, false);
+  }
+
+  componentWillUpdate() {
+    const {data, pluginPurpose} = this.props;
+    let {comments} = this.props;
+    const self = this;
```
tuomas777: why is this (hehe) needed?

Rikuoja: Hm, that's a good question. A remnant of copypaste, don't know why the original plugin had it either. :)

PR #235, opened 2016-07-08
Refactor Hearing into separate component and view

JsseL:
Split views/Hearing/index.js into separate view and component. This
way reusing the Hearing becomes easier and we can reuse the Hearing
component in other views for example.

This work makes admin UI implementation easier.


JsseL:
Getting this merged is quite important due to following reasons:
- This refactoring is quite necessary if we want to implement the hearing admin as planned
- Postponing merge of this would mean solving conflicts every time anything changes in hearing view (during the admin UI development)

Rikuoja:
LGTM!

PR #236, opened 2016-07-08
Hearing manager

JsseL:
## WIP: No need to merge yet

Made this PR for showing the current state of the admin UI development and to list next steps required to finish this thing.
### About the architecture (might still change)
- `HearingManagementView` Renders the `HearingEditor` and hearing preview (`Hearing`). HearingManagementView either initializes a new hearing or fetches existing hearing from API depending whether we are editing or creating a new hearing.
  - `HearingEditor` Dispatches actions made from hearing form or toolbar.
    - `HearingForm` handles the form level stuff (fields, file drops etc.) Form's subsections are all individual components that handle specifics of given sections.
    - `HearingToolbar` provides hearing related actions such as saving and publishing the hearing
### Still todo
- [x] Add sub section management. Currently it's possible to only manage the main section and the closure info section. Sub section management design can be found from the UI proto (see the PR #209) pages 2.1 and 2.2. This should be fairly straight forward by reusing the current `HearingFormStep2` since the subsections have the same fields etc.
- [x] Implement save, publish etc. actions. The current implementation doesn't send anything trough the wire. It has dummy functions for those actions which output the edited hearing into console. This requires also supporting changes to the backend.
- [ ] Add rich text editor for the rich text fields. [Alloy Editor](http://alloyeditor.com/demo/) could be good candidate for this. Currently the fields are plain text fields.
- [ ] Add proper validation/error messages into the form according to backend validation. Simple front end validation would be nice for usability too.
- [x] Add simple validation error messages
- [x] Fetch options for tags, contacts etc from backend. Current implementation uses just dummy data.

admin_ui

**Improvement ideas**
The actual hearing below the modal changes in real time during editing. Though most of it is not visible since the modal blocks visibility. Using a "sidebar editor" instead of modal could make this really nice to use. See the really raw idea tested in the screenshot below.

editor_in_sidebar


src/components/Hearing.js
```diff
@@ -104,6 +113,18 @@ export class Hearing extends React.Component {
     );
   }
 
+  getManageButton() {
```
Rikuoja: Yep. What I'm thinking, should the toolbar actually be visible whenever we are viewing an editable hearing? Currently, clicking "Muokkaa" is required before the user can see the toolbar - is this a good precaution?

src/components/admin/HearingToolbar.js
```diff
@@ -0,0 +1,99 @@
...
+    const actions = [
+      <Button bsStyle="default" onClick={this.props.onEdit} key="edit">
+        <FormattedMessage id="editHearing"/>
+      </Button>
+    ];
+    const open = isOpen(hearing);
+    if (open && hearing.published) {
+      statusLabel = <FormattedMessage id="published"/>;
+      actions.push(
+        <Button bsStyle="danger" onClick={this.props.onCloseHearing} key="close">
```
Rikuoja: Yep, in general they should not be unpublished. However, sometimes there is actually stuff that should not have been published in the first place, and is accidentally published. Perhaps allow unpublishing while the Hearing is open, but do not allow it once the Hearing is closed and historical?

Rikuoja: Another option is to keep this as is, because unpublishing is a rare occurrence, and perhaps should be done via admins. Though closing a Hearing prematurely is a very rare need indeed, so I'm not sure if we should have a button for that either. Hmm.

JsseL: Maybe a teeny weeny tiny button? 😃

Rikuoja: At least they should not be dramatic and easily available? Somehow hide them? :D

src/components/Hearing.js
JsseL: That would actually be much more straight forward. And if we don't need to keep "toolbarless" hearing view at all, we could just use the `HearingView` to handle read, edit, create use cases. That could simplify things in the long run.

Rikuoja:
This is outdated, though might be useful for future reference.

PR #260, opened 2016-10-12
Upgrade react-leaflet from 0.11.7 to ^0.12.3

JsseL:
This gives us some bug fixes, but most notably support for newer
versions of react-leaflet-draw which is required by becoming admin UI.

[Changelog](https://github.com/PaulLeCam/react-leaflet/blob/master/CHANGELOG.md)

From the changelog.

> [BREAKING] Moved Leaflet instances injection from props to context. All components provided by this lib should continue to work as expected, but custom components need to be updated. Read the UPGRADING file for more information.

That change breaks newer versions of `react-leaflet-draw` and that's the main reason for this upgrade.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "react-dom": "^15.1.0",
     "react-helmet": "^2.2.0",
     "react-intl": "2.0.0-rc-1",
-    "react-leaflet": "^0.11.7",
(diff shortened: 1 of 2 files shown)


JsseL:
Overview map and the map on hearing sidebar seems to be working as before this.

PR #262, opened 2016-10-27
Add questionnaire plugin

Rikuoja:
Implements #261


src/actions/index.js
```diff
@@ -78,12 +78,27 @@ export function fetchSectionComments(hearingId, sectionId) {
   };
 }
 
-export function postSectionComment(hearingId, sectionId, text, authorName, pluginData = null, authCode = "") {
+export function postSectionComment(hearingId,
```
akx: Okay, this is starting to have way, way too many arguments, and since JavaScript doesn't have named args, miscalls will be approximately 110% probable.

Maybe turn the signature into `hearingId, sectionId, commentData={}`, and validate the contents of the `commentData` within the function?

src/components/Hearing.js
```diff
@@ -27,14 +27,14 @@ class Hearing extends React.Component {
     const hearingId = this.props.hearingId;
     const {authCode} = this.props.location.query;
     const mainSection = find(this.props.hearing.sections, (section) => section.type === "main");
-    dispatch(postSectionComment(hearingId, mainSection.id, text, authorName, authCode));
+    dispatch(postSectionComment(hearingId, mainSection.id, text, authorName, null, authCode, null, null, null));
   }
 
-  onPostSectionComment(sectionId, text, authorName, pluginData) {
+  onPostSectionComment(sectionId, text, authorName, pluginData, geojson, label, images) {
```
akx: Likewise, this should probably also be refactored to less indexed args and more structured ones.

src/components/plugins/MapQuestionnaire.js
```diff
@@ -0,0 +1,199 @@
...
+import BaseCommentForm from '../BaseCommentForm';
+import CommentDisclaimer from "../CommentDisclaimer";
+import FormControl from 'react-bootstrap/lib/FormControl';
+import FormGroup from 'react-bootstrap/lib/FormGroup';
+import React from 'react';
+import {injectIntl, FormattedMessage} from 'react-intl';
+import {alert} from '../../utils/notify';
+
+
+class MapQuestionnaire extends BaseCommentForm {
```
akx: This is the third copy-paste of the plugin iframe send message listen listener blah blah blah stuff. It should really be refactored into a common base or other helpers.

akx: Yet the other plugins still have the same bits?

Rikuoja: Yes they do, because they are tied to specific source. They won't be reused ever, as the source is not updated.

src/actions/index.js
Rikuoja: Will do. This resulted into miscalls way back when there were less arguments involved.

Issue #145, opened 2016-02-24
Fix and improve UI tests


akx:
Tests are currently broken.

Need to figure out what the current best practices are for testing apps like these, then apply said best practices.

akx:
UI tests were unbroken by #155, but coverage is still abysmal.

akx:
Tests should probably be switched to use Jest instead of Mocha and Chai, as Jest is designed for testing React apps.

JsseL:
This project indeed would require some better/more tests. Maybe we should put some effort just to bring up the test coverage to reasonable level?

Rikuoja:
Some tests are now present. More are of course always welcome.

PR #264, opened 2016-11-02
Remove Node path hacks and switch to Jest for testing

### LINKED ISSUE #259: Node path hacking should be removed
The UI currently uses hacks to add `app` to the Javascript import path ([`app-module-path` here](https://github.com/City-of-Helsinki/kerrokantasi-ui/blob/master/server/index.js#L5) in the server impl, [webpack configuration](https://github.com/City-of-Helsinki/kerrokantasi-ui/blob/master/conf/webpack/common.js#L38) for the webpack bundle).

This lets one import modules as if they were absolute from the `app` root, but it's not quite idiomatic Node and causes the test suite to require magic and hacks (https://github.com/City-of-Helsinki/kerrokantasi-ui/commit/6545cfbcfbd73e52608714bedbd63386f868a22f).

This ties into #145, too.

akx:
At a cursory glance, everything seems to be working still :)

---

# Part 1: Paths and Lint (#259)

Remove path hacks and fix eslint nits; see below.

I did the following:

* removed the infrastructure for the path hacks (breaking things)
* upgraded the Eslint tooling to the latest and greatest (things are still broken)
* fixed the imports to be relative (things are working again but Eslint complains)
* fixed Eslint nits or disabled less crucial Eslint rules (things are working but Eslint complains no more)

Still, it would best if there was no need for any `eslint-disable` stanza anywhere :)

# Part 2: Test Jest

Switch to Jest for tests (refs #145)

---

Refs #145
Closes #259


akx:
Feel free to ignore the Coveralls result; as the coverage tooling changes, it's kind of natural for the calculated value to fluctuate.

JsseL:
Not sure why devDependencies needed to be moved. Is eslint for example required in production? Probably not. Did eslint have some good reason to complain about those dependencies?

Otherwise these look ok to me.

akx:
@JsseL Eslint complained as there were conditional requires and so on, so I figured it'd be easier to just move most deps to `dependencies` rather than start annotating the lines, or starting to pick out single packages to move.

Also, I have the feeling the project would not have worked with `npm i --production` before anyway.

JsseL:
Ok. Sounds fine. @Rikuoja to me this PR seems to make the code base more up to date and cleaner. So I suggest we get this merged. Due to poor test coverage there is of course the risk of having problems, but I think that risk is kinda small in this case.

PR #265, opened 2016-11-04
Fullscreen map

JsseL:
fullscreen_map
Fullscreen map (seems to have unnecessary scrollbar still)

to_fullscreen
Sidebar "to fullscreen" button

### DIFF
--- a/src/components/FullscreenHearing.js
+++ b/src/components/FullscreenHearing.js
@@ -1,0 +1,45 @@
+import React from 'react';
+import {intlShape} from 'react-intl';
+import FullscreenPlugin from './FullscreenPlugin';
+import {getHearingURL, getMainSection} from '../utils/hearing';
+import {Hearing, wrapHearingComponent} from './Hearing';
+
+
+class FullscreenHearing extends Hearing {
+
+  render() {
+    const hearing = this.props.hearing;
+    const user = this.props.user;
+    const onPostVote = this.onVoteComment.bind(this);
+    const mainSection = getMainSection(hearing);
+
+    return (
+      <div id="hearing">
+        {mainSection ? <FullscreenPlugin
+          canComment={this.isMainSectionCommentable}
+          canVote={this.isMainSectionVotable}
+          comments={this.props.sectionComments[mainSection.id]}
+          detailURL={getHearingURL(hearing, {fullscreen: false})}
+          headerTitle={hearing.title}
+          loadSectionComments={this.loadSectionComments.bind(this)}
+          onPostComment={this.onPostSectionComment.bind(this)}
+          onPostVote={onPostVote}
+          section={mainSection}
+          user={user}
+        /> : null}
+      </div>
+    );
+  }
+}
+
+FullscreenHearing.propTypes = {
+  dispatch: React.PropTypes.func,
+  hearing: React.PropTypes.object,
+  hearingId: React.PropTypes.string,
+  intl: intlShape.isRequired,
+  location: React.PropTypes.object,
+  sectionComments: React.PropTypes.object,
+  user: React.PropTypes.object,
+};
+
+export default wrapHearingComponent(FullscreenHearing);
--- a/src/components/Hearing.js
+++ b/src/components/Hearing.js
@@ -35,37 +35,37 @@ export class Hearing extends React.Component {
 
   onPostHearingComment(text, authorName) {
     const {dispatch} = this.props;
-    const hearingId = this.props.hearingId;
+    const hearingSlug = this.props.hearingSlug;
     const {authCode} = this.props.location.query;
     const mainSection = find(this.props.hearing.sections, (section) => section.type === "main");
     const commentData = {text, authorName, pluginData: null, authCode, geojson: null, label: null, images: []};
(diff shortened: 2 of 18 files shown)


Rikuoja:
Looking good! Fix notifications and posted comment box to render on top of the plugin, currently they are displayed underneath it.

Rikuoja:
On the non-fullscreen page, in place of the plugin introduce a big "Open questionnaire" button or similar. Disable rendering the plugin and comments on the non-fullscreen page, as the user experience is in my opinion suboptimal.

Rikuoja:
Change fullscreen mode top text link to take the user to the non-fullscreen page. The headline should be the name of the hearing, shortened if it does not fit on mobile?

Rikuoja:
Perhaps introduce an arrow or divider element between Helsinki coat of arms and the Hearing headline. This makes it evident that the Hearing headline is just for the single hearing, while the coat of arms will take us to kerrokantasi front page.

src/components/Hearing.js
```diff
@@ -167,21 +190,26 @@ class Hearing extends React.Component {
...
-                onPostComment={this.onPostHearingComment.bind(this)}
-                canVote={mainSectionVotable}
-                onPostVote={onPostVote}
-                canSetNickname={user === null}
-              />
-            </div>
-            <hr/>
-            <a href={reportUrl}><FormattedMessage id="downloadReport"/></a>
+            {fullscreenMapPlugin ?
+              null :
```
JsseL: Not sure how far this logic can stretch, but for now we just blatantly omit comment listing if Hearing has fullscreenMap. Not sure if we wan't to have fullscreen maps in the future and still have comments.

src/utils/hearing.js
```diff
@@ -0,0 +1,32 @@
...
+export function getMainSection(hearing) {
+  return _.find(hearing.sections, (section) => section.type === "main");
+}
+
+/*
+* Return URL to hearing view. Accepts optional fullscreen parameter
+* to force fullscreen query parameter.
+ */
+export function getHearingURL(hearing, {fullscreen} = {}) {
+  const url = `/hearing/${hearing.id}`;
```
Rikuoja: This should probably to be fixed to `/hearing/${hearing.id}`instead? Currently, redirect from hearing/slug to hearing/slug?fullscreen=true does not work, because we are making the assumption that the original url was of the form hearing/id.

Also, this fix would have the admittedly handy side effect of fixing issue #247 =)

src/i18n/fi.json
```diff
@@ -50,5 +50,6 @@
   "publishingQueue": "Julkaisujono",
   "drafts": "Luonnokset",
   "nickname": "Nimimerkki",
-  "registered": "rekisteröitynyt"
+  "registered": "rekisteröitynyt",
+  "openFullscreenMap": "Avaa karttalla"
```
Rikuoja: Oh and should be "kartalla", double 'tt' is subject to consonant gradation when inflecting =)

JsseL: 🙈

src/utils/hearing.js
JsseL: Cool. Didn't know backend already returns the same response with `/v1/hearing/<slug>` request. Will change. I'll double check that all hearing links also use this function so that we can close #247 as a side effect.

Didn't manage to reproduce the redirection problem you mentioned. `/hearing/slug` does indeed redirect to `hearing/slug?fullscreen=true` when hearing should default to fullscreen.

JsseL: Change it now to use `hearing.slug` instead of `id`. Also refactored `OverviewMap` so that it uses this same function. I think this should now solve #247 altogether.

src/components/FullscreenHearing.js
```diff
@@ -0,0 +1,53 @@
...
+import {Hearing, wrapHearingComponent} from './Hearing';
+
+
+class FullscreenHearing extends Hearing {
+
+  render() {
+    const hearing = this.props.hearing;
+    const user = this.props.user;
+
+    const hearingAllowsComments = !hearing.closed && (new Date() < new Date(hearing.close_at));
```
stephawe: It might seem a little overkill but I would say that this could be a function in the Hearing. Mainly to avoid having several places to change if the logic changes. This basically applies to the following as well `mainSection`, `mainSectionCommentable` and possibly even `mainSectionVotable`.

In my opinion, since this class is only to provide `fullscreen`, it should not be possible that at some point, its behavior differs from the regular hearing. So, if possible, it is nicer to push all the logic to the Hearing class.

Rikuoja: An excellent point. As we have had to edit the logic for these booleans in a number of occasions, I think the logic should not be duplicated.

JsseL: Well spotted. I actually was going to do this, but felt the urge to rush this PR. Will refactor right away!

Rikuoja:
Oh and yet another finishing touch: currently the sidebar features a link "Mielipiteet" even though the section is missing, so the link only scrolls to the button saying "Avaa kartalla". Instead, how about reintroducing the link directly the full screen map in the Sidebar Contents, as well? Can't go wrong with multiple buttons taking us to the same full screen map, can we? :D

PR #281, opened 2016-12-12
Enable switching language

vaaralav:
**Not ready yet!**

Fixes #230.

- Add language switcher
- Add support for translating API entities. **For testing purposes appends language code to the string!**

Missing:

- [ ] Swedish i18n translations, `sv.json`
- [ ] English i18n translations, `en.json`
- [ ] API support for translated values
- [ ] Change hearing language without changing UI language
- [x] Don'tappend language code to translated strings

PR #282, opened 2016-12-12
Subsection links to sidebar

### LINKED ISSUE #243: Display section headlines below sections in the left bar, to make navigating within a hearing easier?

vaaralav:
Fixes #243 .
~Depends on PR #281 .~

- [x] Sidebar scrolling problems when sidebar is taller than viewport
- [x] Fixed nav covers elements with `#` linking
- [ ] Don't scroll sidebar separately, but keep sidebar always visible.
- [ ] Limit sidebar height only when footer is visible


vaaralav:
- [ ] Sidebar scrolling problems when sidebar is taller than viewport
- [ ] Fixed nav covers elements with `#` linking

Rikuoja:
Currently, the sidebar displayed is actually not as tall as the page itself. Ideally, the sidebar would extend to the bottom of the screen as it currently does. 
image

Rikuoja:
In addition, scrolling the sidebar separately from the main page results in rather funny behavior, in which only top of the sidebar may be visible even if we have scrolled the main page downwards. Similarly, only the bottom of the sidebar may be visible at the top of the page, if we have hit the sidebar while scrolling up and down.

image

Ideally, the sidebar would scroll up and down along with the page itself, but stop scrolling once we have hit the bottom of the sidebar. I couldn't figure out if it's possible to do this with css.

vaaralav:
Decided to do #243 and #289 together: https://github.com/vaaralav/kerrokantasi-ui/tree/feature/%23289-%23243-refactor-hearing-navigation

PR #283, opened 2016-12-13
Comment list sorting and pagination

### LINKED ISSUE #175: Comment order by 1) most popular, 2) created by as tiebreaker

vaaralav:
Fixes #175 .
TODO:

- [x] Enable commenting with SortableCommentList


Rikuoja:
The API already offers comment pagination at the root `/comment/?section=foo` endpoint. Also, ordering by `n_votes`, `created_at` etc. will soon be implemented.

Also, ordering by `created_at` could be used as an alternative to `n_votes` in the UI, so that when the user POSTs their own comment, they will see the posted comment first in the list, like they currently do? Of course, this would require a UI element for changing comment ordering, with `n_votes` being the default, and reloading the comment list whenever the user changes ordering?

Let's transition to using that endpoint for fetching the first comments when loading each section, then loading more when we scroll to the bottom of the comments. Autoload when scrolling to the bottom, however, will have to wait until the sections are actually displayed on a separate page each.

jonnetanninen:
Is this now working as supposed to?

assets/less/hel/commentlist.less
```diff
@@ -0,0 +1,129 @@
...
+          padding-top: 10px;
+      }
+  }
+
+  .comment-conditions {
+      padding-top: 15px;
+      font-size: 1.4rem;
+  }
+
+  .section-title {
```
Rikuoja: For some reason, this doesn't seem to be working. At least the cursor is wrong and the # comments doesn't float right.

assets/less/hel/hearing-page.less
```diff
@@ -225,7 +112,6 @@
         font-size: 20px;
     }
     .commentlist {
-        background-color: @brand-gray-light;
         padding: 15px;
```
Rikuoja: Here, I would restore the gray background for the section comments until each section is on a separate page. Without the background there is no visual distinction between section comments and hearing comments that follow them.

src/components/SortableCommentList.js
```diff
@@ -0,0 +1,207 @@
...
+              section={section}
+              comments={sectionComments.results}
+              totalCount={sectionComments.count}
+              onPostComment={this.onPostHearingComment.bind(this)}
+              onPostVote={this.onVoteComment.bind(this)}
+              isLoading={this.state.showLoader}
+              {...rest}
+            />
+            <p className="sortable-comment-list__count">
+              <FormattedMessage id="commentsOutOf" values={{current: sectionComments.results.length, total: sectionComments.count }}/>
```
Rikuoja: Here, scrolling to the bottom increases the total count beyond those given in the section. 
image

src/components/SortableCommentList.js
```diff
@@ -0,0 +1,207 @@
...
+            </p>
+            <Waypoint onEnter={this.handleReachBottom}/>
+          </div>
+        }
+        {this.state.showLoader ?
+          <div className="sortable-comment-list__loader">
+            <LoadSpinner />
```
Rikuoja: As in the above picture, the spinner is for some reason still present and animating even though all the comments (and more :) have been loaded.

src/components/SortableCommentList.js
```diff
@@ -0,0 +1,207 @@
...
+                pluginPurpose="viewHeatmap"
+                comments={sectionComments.results}
+              />
+              <div className="image-caption">Siirrettyjen ja lisättyjen asemien tiheyskartta.</div>
+            </div>
+            : null
+        }
+        <form className="sort-selector">
+          <FormGroup controlId="sort-select">
+            <ControlLabel><FormattedMessage id="commentOrder"/></ControlLabel>
```
Rikuoja: I would move the ordering control to the right hand side (and make it narrower) so it is more in line with location of the ordering control in the hearing search page.

Also, this allows the ordering control to be on the same row as the Answer button when the comment input box is closed. This means less vertical whitespace and more comments visible for the user.

Rikuoja:
If you cannot reproduce any of the bugs listed above, they are most probably caused by a botched merge on my part.

In such a case, feel free to ignore said bug report and rebase :)

jonnetanninen:
Fixed the count bug which was real, others couldn't reproduce

Rikuoja:
Great, then just a rebase is needed!

jonnetanninen:
Rebased and should be ready to go

Rikuoja:
I'm still getting the bug, apparently it only appears in subsection comments. In a simple Hearing, everything works fine.

If a subsection is expanded, its comment list is buggy, and actually the original Hearing comments disappear completely when the subsection is expanded!

image

Rikuoja:
Incidentally, this is why I the subsection `background-color: @brand-gray-light;` should be restored.

Otherwise, you see two comment lists with no visual distinction, which is very confusing.

Rikuoja:
Even more weirdly, I am only getting the problem in the last of my subsections. The other subsections work fine.

Rikuoja:
It would seem I get the bug whenever the subsection has commenting disabled. In the subsections where commenting is allowed, everything works fine.

Rikuoja:
Oh great. Further testing indicates it's not about commenting.

It's in a subsection with less than ten comments, where no further comments can be loaded. That is most probably the cause.

Rikuoja:
Okay, nailed it :D If a subsection (not hearing) has less than ten comments, it will display the comments twice and still believes it is loading more.

image

Rikuoja:
Oh and let's move the ordering filter menu to the right. It shouldn't take up a complete row, because the text is not that long. This way, "Vastaa" button (if present) can be on the same row as the ordering filter. The result is less whitespace and more comments visible to the user. 
image

jonnetanninen:
Section comments are fixed now, it was a bug in the reducer.

Putting the button and filter inline would take excessive refactoring, estimation is from 4 to 8 hours, would you like it to be done regardless?

assets/less/hel/hearing-page.less
```diff
@@ -228,7 +228,6 @@
         font-size: 20px;
     }
     .commentlist {
```
Rikuoja: Oh, the css implementation requires moving the sort-selector after the CommentForm element, but no other changes needed.

Rikuoja:
Oh, and the css implementation requires moving the sort-selector after the CommentForm element, but no other changes needed.

jonnetanninen:
CSS changes implemented

PR #284, opened 2016-12-14
Feature/image upload

### LINKED ISSUE #250: Allow the user to add image to their comment

jonnetanninen:
Add feature for adding images to comments, fixes #250

### DIFF
--- a/src/components/BaseCommentForm.js
+++ b/src/components/BaseCommentForm.js
@@ -8,13 +8,11 @@ import Icon from '../utils/Icon';
 import {getImageAsBase64Promise} from '../utils/hearing';
 import CommentDisclaimer from './CommentDisclaimer';
 import forEach from 'lodash/forEach';
-import round from 'lodash/round';
 
 class BaseCommentForm extends React.Component {
   constructor(props, context) {
     super(props, context);
-    this.state = {collapsed: true, commentText: "", nickname: "", imageTooBig: false};
-    this.imageTooBig = this.imageTooBig.bind(this);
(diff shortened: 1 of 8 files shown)


src/components/BaseCommentForm.js
```diff
@@ -109,6 +151,36 @@ class BaseCommentForm extends React.Component {
...
+                {this.state.imageTooBig
+                  ? <div className="comment-form__image-too-big">
+                    <FormattedMessage id="image_too_big"/>
+                  </div>
+                  : null}
+                {this.refs.images
+                  ? this.getSelectedImagesAsArray(this.refs.images.files).map(
+                      (image) =>
+                        <p style={image.size > 1000000 ? {color: 'red'} : null}>
+                          {image.name} {round((image.size / 1000000), 3) + 'MB'}
```
Rikuoja: Wonder if we should actually display the image thumbnails instead of the image names here? So the user will see their comment the same way as after it is posted? Also, I would render the thumbnails next to the button (not below), so the form does not get any taller than necessary.

jonnetanninen: How about now?

Rikuoja: Looks good! Though we probably should add a help text that tells the user how to select several images at once in the dialog (press Ctrl/Cmd/whatever), otherwise they will end up clicking the button again to add more images and wonder why the previously selected image was replaced (as I did at first).

jonnetanninen: Added helper text

PR #285, opened 2016-12-16
Feature/#195 contact person component

### LINKED ISSUE #195: make a new UI element for the public servant contact info
the contact info of the people at the city is standard content for each hearing. should we move this to the left column in a separate element? on a small screen it could drop to the very bottom of the screen below the comments?

image

vaaralav:
Fixes #195 .

There's some scrolling problems with sidebar being taller but that should get fixed with PR #282 .

### DIFF
--- a/src/components/ContactCard.js
+++ b/src/components/ContactCard.js
@@ -3,7 +3,9 @@ import {FormattedMessage} from 'react-intl';
 
 const ContactMethod = ({name = null, value, href}) =>
   <div className="contact-card__contact-method">
-    {name} {href ?
+    {name}
+    {name && ': '}
+    {href ?
       <a href={href}>{value}</a>
       : value
     }
--- a/src/components/ContactCard.js
+++ b/src/components/ContactCard.js
@@ -1,0 +1,26 @@
+import React, {PropTypes} from 'react';
+
+const ContactCard = ({className = '', name, title, phone, email, organization, ...rest}) =>
+  <div className={`contact-card ${className}`} {...rest}>
+    <header>
+      {name && <h4>{name}</h4>}
+      {title || organization ?
+        <h6>{title || ''}{title && organization ? `, ${organization}` : ''}</h6>
+        : null}
+    </header>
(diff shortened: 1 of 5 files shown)


src/views/Hearing/Sidebar.js
```diff
@@ -85,6 +86,10 @@ class Sidebar extends React.Component {
               <SocialBar />
               {hearingMap}
             </Col>
+            {hearing.contact_persons && hearing.contact_persons.map((person, index) =>
```
Rikuoja: I would add a header LISÄTIEDOT with the same style as other sidebar headers.

src/components/ContactCard.js
```diff
@@ -0,0 +1,51 @@
...
+};
+
+const contactCardStyles = {
+  padding: '10px 0'
+};
+
+const ContactCard = ({className = '', name, title, phone, email, organization, ...rest}) =>
+  <div className={`contact-card ${className}`} style={contactCardStyles} {...rest}>
+    <header>
+      {name && <h5>{name}</h5>}
```
Rikuoja: Perhaps change to a smaller font and try out grey text for some lines, so the style would be more in line with other sidebar features and not that prominent? Don't know what would be the optimal size here.

src/components/ContactCard.js
```diff
@@ -0,0 +1,51 @@
...
+      {title || organization ?
+        <strong>
+          {title || ''}
+          {title && organization ? <br/> : null}
+          {organization || ''}
+        </strong>
+        : null}
+    </header>
+    <div className="contact-card__information">
+      {phone && <ContactMethod name={<FormattedMessage id="phone"/>} value={phone} href={`tel:${phone}`}/>}
```
Rikuoja: Add phone and mail icons, to be more in line with other sidebar elements with icons.

src/components/ContactCard.js
Rikuoja: Oh, and perhaps even a person/bureaucrat icon here next to the name or title?

PR #291, opened 2016-12-21
Feature/hearing list search

### LINKED ISSUE #255: Allow sorting, filtering, search in Hearing list page

jonnetanninen:
Fixes #255

### DIFF
--- a/src/components/HearingList.js
+++ b/src/components/HearingList.js
@@ -97,10 +97,10 @@ class HearingList extends React.Component {
 
     return (
       <div>
+        <HearingsSearch handleSearch={handleSearch} labels={labels} handleLabelSearch={handleLabelSearch}/>
+        <HearingListTabs/>
         {isLoading && <LoadSpinner />}
         <div className={`hearing-list${isLoading ? '-hidden' : ''}`}>
-          <HearingsSearch handleSearch={handleSearch} labels={labels} handleLabelSearch={handleLabelSearch}/>
-          <HearingListTabs/>
           <HearingListFilters labels={labels} handleChangeFilter={handleChangeFilter} handleSort={handleSort}/>
           {hearings.map(
           (hearing) => <HearingListItem hearing={hearing} key={hearing.id}/>
--- a/src/components/HearingList.js
+++ b/src/components/HearingList.js
@@ -17,7 +17,7 @@ import getAttr from '../utils/getAttr';
 import HearingsSearch from './HearingsSearch';
 
 const HearingListTabs = () =>
-  <Nav className="hearing-list__tabs" bsStyle="tabs" activeKey="1" pullRight={1}>
+  <Nav className="hearing-list__tabs" bsStyle="tabs" activeKey="1">
     <NavItem eventKey="3" disabled className="hearing-list__tabs-empty"/>
     <NavItem eventKey="1" href="/home">Lista</NavItem>
     <NavItem eventKey="2" title="Item">Kartta</NavItem>
@@ -26,7 +26,6 @@ const HearingListTabs = () =>
 const HearingListFilters = ({handleSort}) =>
   <div className="hearing-list__filter-bar">
     <FormGroup controlId="formControlsSelect" className="hearing-list__filter-bar-filter">
-      <ControlLabel><FormattedMessage id="sort"/></ControlLabel>
       <FormControl componentClass="select" placeholder="select" onChange={(event) => handleSort(event.target.value)}>
         <option value="-created_at"><FormattedMessage id="newestFirst"/></option>
         <option value="created_at"><FormattedMessage id="oldestFirst"/></option>
@@ -38,6 +37,7 @@ const HearingListFilters = ({handleSort}) =>
         <option value="n_comments"><FormattedMessage id="leastCommented"/></option>
       </FormControl>
     </FormGroup>
+    <ControlLabel className="hearing-list__filter-bar-label"><FormattedMessage id="sort"/></ControlLabel>
   </div>;
 
 HearingListFilters.propTypes = {
--- a/src/components/HearingList.js
+++ b/src/components/HearingList.js
@@ -4,6 +4,7 @@ import Nav from 'react-bootstrap/lib/Nav';
 import NavItem from 'react-bootstrap/lib/NavItem';
 import FormGroup from 'react-bootstrap/lib/FormGroup';
 import FormControl from 'react-bootstrap/lib/FormControl';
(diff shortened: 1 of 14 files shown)


src/components/HearingList.js
```diff
@@ -9,6 +14,47 @@ import LabelList from './LabelList';
...
+        <option value="close_at"><FormattedMessage id="firstClosing"/></option>
+        <option value="-open_at"><FormattedMessage id="lastOpen"/></option>
+        <option value="open_at"><FormattedMessage id="firstOpen"/></option>
+        <option value="-n_comments"><FormattedMessage id="mostCommented"/></option>
+        <option value="n_comments"><FormattedMessage id="leastCommented"/></option>
+      </FormControl>
+    </FormGroup>
+    {labels.length !== 0 && <FormGroup controlId="formControlsSelect" className="hearing-list__filter-bar-filter">
+      <ControlLabel><FormattedMessage id="chooseTopic"/></ControlLabel>
+      <FormControl componentClass="select" placeholder="select" onChange={(event) => handleChangeFilter(event.target.value)}>
```
Rikuoja: I have the feeling drop-down menu doesn't really cut it in here. There are too many possible labels to select from, and there will be even more in the future.

Currently there is no autocomplete API endpoint for fetching labels when typing (though there could be), but how about a text box (below the title search text box) that allows you to start typing and select matching label from the autocomplete menu? This would allow you to select several labels.

src/components/HearingList.js
```diff
@@ -57,16 +103,30 @@ HearingListItem.contextTypes = {
...
-    const {state, data} = (this.props.hearings || {});
-    if (state !== "done") return <LoadSpinner />;
-    return (<div className="hearing-list">{data.map(
-      (hearing) => <HearingListItem hearing={hearing} key={hearing.id}/>
-    )}</div>);
+    const {hearings, isLoading, labels, handleChangeFilter, handleSort, handleSearch} = this.props;
+
+    return (
+      <div>
+        {isLoading && <LoadSpinner />}
```
Rikuoja: I would move the LoadSpinner below the tabs, so that the search boxes are visible all the time. Only the list would be updated (displaying the spinner) when something is typed in the search boxes.

Rikuoja:
I would suggest some layout changes (pictured) to the hearing list page, so the hearing list itself would be more prominent and the UI elements would take up less vertical space. Currently, the list is barely visible on a small screen because of the large UI elements?

image

src/components/HearingList.js
Rikuoja: Edit: I added the API filter for label searching. So you could use `/v1/label/?label=text` to search and autocomplete labels, matching the text typed by user.

jonnetanninen: The whitespace is obvious one and easy to fix, but I have few questions about the "Move label search below topic search, so they will have common search box" from the picture:

It is not actually a search now, it's a filter. After thinking this through with Ville, it made more sense this way than the way it was in the original wireframes. The way it is implemented now is way way easier (read: less expensive) to implement and better value & less confusing for the user.

The problem with the original wireframe was, that the logic how it should was not clear at all. What if the user fills both search fields? Should it show hearings that match the Title AND Label or Title OR Label?

The other problem was, how the user knows which labels there actually are? Now this should be a less of a problem with the autocomplete endpoint (which did not exist at the time of first iteration), but it won't solve it completely. The user still has to make a guess instead of skimming the available options to find something interesting.

TL;DR:

What I suggest that we do, is keep the label thing logic as a filter like it is now. If the dropdown size is a problem, it can be changed in to a multiselect box as you mentioned earlier. And to prevent the user confusing it with the search keep it where it is.

If this is to be changed the way suggested in the picture I need the following to implement:

1. Specification for the logic the search should work:
    a. Should the hearings be shown with AND or OR logic if both fields are selected? 
    b. Should the filtering happen onChange or onSubmit (the former was very clunky when I tested with the Title search)?

2. Specification how the label search logic should work:
    a. When the customer writes to the search field, should API call be made onChange and the suggestions be shown for the customer?
    b. What happens when the customer selects a label, should the filtering happen then or onSubmit?
    c. Should the customer be able to select multiple labels? If yes, how and where should the selected labes be rendered?
    d. If multiple labels are selected, should they work with AND or OR logic?

Rikuoja: Hm, essentially text search and label search are both "filters" and as such could be treated equally. However, it might very well be that the user prefers to search by text (and not label), so the text search box could have primacy and the label could be shown as an additional filter. To me, though, it is more confusing have the label filter (which behaves like the text filter) next to the order control (which does not filter at all).

Any way, the location of the filter control shouldn't really affect the behavior of the filter:

1. a) In any case, the behavior should be AND (if the user has specified both text and Label filters), obtained with the backend `v1/hearing/?title=abc&label=1,3` query. Obviously, in most cases the user wouldn't fill both boxes simultaneously.
b) onChange would be preferable. However, it might be that the contains filter for the text in the backend is too slow, so onSubmit can be used if it feels more responsive.

2. a) The label autocomplete should (hopefully) work fast enough for onChange filtering of labels based on typing, so yes :)
b) onChange would be preferable but, again, if it's not responsive enough we'll have to try onSubmit instead.
c) I would suggest a multiple select field like https://github.com/JedWatson/react-select with `multi={true}`, or something similar.
d) In the backend multiple labels currently work with OR logic. Admittedly, it is a bit weird that a single filter works with OR logic and multiple filters combine with AND logic, but actually that seems to be our standard in endpoints which have multiple possible filters, so let's do that for now and see how it feels in practice.

jonnetanninen:
Requested changes implemented

Rikuoja:
I would actually move the Label search box next to the text search box:
```
.hearings-search {
    &__container {
        padding-bottom: 10px;
        height: 80px;
    }

    &__label {
        width: 48%;
        float: right;
    }

    &__text {
        width: 48%;
        display: inline-block;
    }

    &__button {
        float: left;
    }
}
```
This way they take minimal vertical space, though the latter box will then expand vertically on very small screens if many labels are added. I think this is better behavior.

Also, smaller font, radically less whitespace and `float:right` for the `hearing-list__filter-bar`. Also: put the "Järjestä" text on same row with the FormControl itself, so it takes less vertical space.

The goal is that the search results would always be visible when changing the search parameters, so we want to compress the top elements vertically. Currently, there is so much vertical whitespace that the user has to scroll to check the results after changing search parameters.

jonnetanninen:
Style changes implemented

Rikuoja:
LGTM! Just a rebase needed.

jonnetanninen:
Rebased

PR #299, opened 2017-01-13
Feature/#295 display map

### LINKED ISSUE #295: Display map with hearings on front page and hearings search page

jonnetanninen:
Fix #295

### DIFF
--- a/src/components/HearingList.js
+++ b/src/components/HearingList.js
@@ -113,14 +113,14 @@ HearingListItem.propTypes = {
 class HearingList extends React.Component {
 
   render() {
-    const {hearings, isLoading, labels, handleSort, handleSearch, handleLabelSearch, language, handleChangeTab, activeTab, hearingMap, showOnlyOpen, toggleShowOnlyOpen} = this.props;
-    console.log(hearingMap);
-    const hearingListMap = (hearingMap && hearingMap.data ? (<div className="hearing-list-map map">
+    const {hearings, isLoading, labels, handleSort, handleSearch, handleLabelSearch, language, handleChangeTab, activeTab, showOnlyOpen, toggleShowOnlyOpen} = this.props;
+    const hearingsToShow = !showOnlyOpen ? hearings : hearings.filter((hearing) => !hearing.closed);
+    const hearingListMap = (hearingsToShow ? (<div className="hearing-list-map map">
       <HearingListFilters handleSort={handleSort}/>
       <Checkbox inline readOnly checked={showOnlyOpen} onChange={() => toggleShowOnlyOpen()}>
         <FormattedMessage id="showOnlyOpen"/>
       </Checkbox>
-      <OverviewMap hearings={hearingMap.data} style={{width: '100%', height: '40%'}} enablePopups />
+      <OverviewMap hearings={hearingsToShow} style={{width: '100%', height: '40%'}} enablePopups />
     </div>) : null);
 
     return (
@@ -136,7 +136,7 @@ class HearingList extends React.Component {
             )}
           </div>
         }
-        {activeTab === 'map' && hearingListMap}
+        {activeTab === 'map' && !isLoading && hearingListMap}
       </div>
     );
   }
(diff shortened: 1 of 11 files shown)


Rikuoja:
Let's get rid of the separate Map page, so that the titlebar Map link takes us to the map tab.

In the same vein, let's increase the size of the map in the map tab so that it is as wide as possible and almost screen height. Currently, the map only has the width of the tab, and the height is way too small.

src/components/HearingList.js
```diff
@@ -113,19 +113,19 @@ HearingListItem.propTypes = {
...
   render() {
-    const {hearings, isLoading, labels, handleSort, handleSearch, handleLabelSearch, language, handleChangeTab, activeTab, showOnlyOpen, toggleShowOnlyOpen} = this.props;
+    const {hearings, isLoading, labels, handleSort, handleSearch, handleLabelSearch, language, activeTab, showOnlyOpen, toggleShowOnlyOpen} = this.props;
     const hearingsToShow = !showOnlyOpen ? hearings : hearings.filter((hearing) => !hearing.closed);
     const hearingListMap = (hearingsToShow ? (<div className="hearing-list-map map">
       <Checkbox inline readOnly checked={showOnlyOpen} onChange={() => toggleShowOnlyOpen()} style={{marginBottom: 10}}>
         <FormattedMessage id="showOnlyOpen"/>
       </Checkbox>
-      <OverviewMap hearings={hearingsToShow} style={{width: '100%', height: '40%'}} enablePopups />
+      <OverviewMap hearings={hearingsToShow} style={{width: '130%', height: 600}} enablePopups />
```
Rikuoja: Hm, now the map flows over the edge on small devices. Is it possible to get the width fixed to the screen size (not viewport size) so it would span the screen width? Or would we need to actually make the whole list view (tabs and all) fit the screen width?

src/views/Home.js
```diff
@@ -33,6 +34,11 @@ class Home extends React.Component {
   render() {
     const {formatMessage} = this.props.intl;
     const {topHearing, openHearings, language} = this.props;
+    const hearingMap = (openHearings && openHearings.data ? (<div className="container map">
+      <h4><FormattedMessage id="open-hearings-on-map"/></h4>
+      <OverviewMap hearings={openHearings.data} style={{width: '100%', height: '40%'}} />
```
Rikuoja: Oh, and increase the height to 600px here as well, and the width to full screen width.

src/views/Home.js
```diff
@@ -28,33 +36,56 @@ class Home extends React.Component {
...
-          <div className="feedback-box">
-            <a href="mailto:dev@hel.fi?subject=Kerro kantasi -palaute">
-              <h2 className="feedback-prompt"><FormattedMessage id="feedbackPrompt"/></h2>
-            </a>
-          </div>
+        </Row>
+      </div>
+      <Row>
+        <Col sm={12}>
+          {hearingMap}
```
Rikuoja: The map doesn't display any geometry; we should fetch the open hearings on the front page with `?include=geojson` to get the geometry on the map.

src/views/Home.js
```diff
@@ -28,33 +36,56 @@ class Home extends React.Component {
...
-      <Helmet title={formatMessage({id: 'welcome'})}/>
-      <h1><FormattedMessage id="welcome"/></h1>
-      <p className="lead"><FormattedMessage id="welcomeMessage"/></p>
-      {topHearing && <FullWidthHearing hearing={topHearing}/>}
-      <hr />
-      <Row>
-        {openHearings && !openHearings.isFetching &&
+    const {isMobile} = this.state;
+    const hearingMap = (openHearings && openHearings.data ? (<div className="map">
+      <h4><FormattedMessage id="open-hearings-on-map"/></h4>
```
Rikuoja: Change this class to h2.page-title to visually separate the header

src/views/Home.js
```diff
@@ -28,33 +36,56 @@ class Home extends React.Component {
...
-    return (<div className="container">
-      <Helmet title={formatMessage({id: 'welcome'})}/>
-      <h1><FormattedMessage id="welcome"/></h1>
-      <p className="lead"><FormattedMessage id="welcomeMessage"/></p>
-      {topHearing && <FullWidthHearing hearing={topHearing}/>}
-      <hr />
-      <Row>
-        {openHearings && !openHearings.isFetching &&
+    const {isMobile} = this.state;
+    const hearingMap = (openHearings && openHearings.data ? (<div className="map">
```
Rikuoja: Add the same 15px left and right padding for the hearing map element as in the rest of the page elements.

jonnetanninen:
Suggested changes applied

PR #300, opened 2017-01-16
Feature/#289 question views

### LINKED ISSUE #289: Questionnaire views
Questions should open in their own view instead of expanding/collapsing in the hearing view.

- [x] Enable commenting etc. on section page
- [x] Fix sidebar scrolling. See comments https://github.com/City-of-Helsinki/kerrokantasi-ui/pull/282.
- [ ] Highlight active section in sidebar. See #256 . (Optional for this issue, but is partly done just by using `<Link>` instead of `<a>`)

jonnetanninen:
Fix #289

### DIFF
--- a/assets/less/hel/hearing-page.less
+++ b/assets/less/hel/hearing-page.less
@@ -3,6 +3,7 @@
     padding: 15px;
     font-size: 1.4rem;
     margin-bottom: 25px;
+    overflow-y: scroll;
 
     .language-link {
         padding: 4px;
--- a/assets/less/hel/hearing-page.less
+++ b/assets/less/hel/hearing-page.less
@@ -3,8 +3,6 @@
     padding: 15px;
     font-size: 1.4rem;
     margin-bottom: 25px;
-    overflow-y: scroll;
-    max-height: 85%;
 
     .language-link {
         padding: 4px;
@@ -236,16 +234,14 @@
     h2:first-letter {
         text-transform:capitalize;
     }
-
-    .commenticon {
-        float: right;
-    }
 }
 
 .hearing-section {
-    h2 {
-        font-size: 20px;
+
+    .sortable-comment-list {
+        padding-top: 50px;
     }
+
     .commentlist {
         padding: 15px;
     }
@@ -263,4 +259,9 @@
   background: @brand-gray-light;
   padding: 20px 15px 15px;
   text-align: center;
+}
+
+
+.commenticon {
+    float: right;
 }
--- a/assets/less/hel/hearing-page.less
+++ b/assets/less/hel/hearing-page.less
@@ -227,7 +227,8 @@
 }
 
 #hearing-comments,
-[id^='hearing-sectiongroup-'] {
+[id^='hearing-sectiongroup-'],
+.commentlist {
     padding-top: 50px;
 
     h2:first-letter {
--- a/assets/less/hel/hearing-page.less
+++ b/assets/less/hel/hearing-page.less
@@ -5,6 +5,32 @@
     margin-bottom: 25px;
     overflow-y: scroll;
 
+    .list-group {
+        .list-group-item {
+            color: @gray-light;
+            &:hover {
+                color: #FFFFFF;
+                background: #005eb8;
+            }
+
+            .active {
+                color: #FFFFFF;
+            }
+        }
+        a {
+
+            :hover {
+                color: #FFFFFF;
+            }
+            .active {
+                color: #FFFFFF;
+            }
+        }
+        .active {
+            color: #FFFFFF;
+        }
+    }
+
     .language-link {
         padding: 4px;
         border-radius: 4px;
--- a/assets/less/hel/hearing-page.less
+++ b/assets/less/hel/hearing-page.less
@@ -3,6 +3,8 @@
     padding: 15px;
     font-size: 1.4rem;
     margin-bottom: 25px;
+    overflow-y: scroll;
+    max-height: 85%;
 
     .language-link {
         padding: 4px;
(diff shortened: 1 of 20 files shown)


assets/less/hel/hearing-page.less
```diff
@@ -3,6 +3,8 @@
     padding: 15px;
     font-size: 1.4rem;
     margin-bottom: 25px;
+    overflow-y: scroll;
```
Rikuoja: Let's lock the sidebar scrolling to the main text. At the moment, nobody realizes the sidebar is scrollable, so the rest of the sidebar will remain unseen by users.

assets/less/hel/question-page.less
```diff
@@ -0,0 +1,35 @@
...
+      margin: auto 5px;
+      padding: 5px;
+      border: 1px solid @link-color;
+
+      .type-name {
+        text-transform: lowercase;
+      }
+
+      &:hover {
+        border-color: @link-hover-color;
```
Rikuoja: Let's change the "currently viewed" color to grey, while keeping the hover color purple. This is more in line with the currently selected language color in the sidebar (also grey).
EDIT: put the comment in the wrong place. I mean the sidebar "currently viewed" section color.

src/views/Hearing/Sidebar.js
```diff
@@ -102,17 +104,29 @@ class Sidebar extends React.Component {
               <div className="sidebar-section contents">
                 <h4><FormattedMessage id="table-of-content"/></h4>
                 <ListGroup>
-                  <ListGroupItem href="#hearing">
+                  {!isQuestionView && <ListGroupItem className={currentlyViewed === '#hearing' && 'active'} href="#hearing">
```
Rikuoja: Let's display the whole Hearing contents (and all subsections) in the sidebar at all times, instead of changing the sidebar when we go to the subsections.

That is, we should have the subsections listed in the sidebar below the Vaihtoehdot header (even if we are looking at the hearing main page). Similarly, the whole hearing contents should be visible at all times (even if we are looking at a subsection), with the current subsection highlighted.

src/components/SectionContainer.js
```diff
@@ -0,0 +1,346 @@
...
+        </h1>
+
+        <Row>
+          <Sidebar currentlyViewed={'#hearing-sectiongroup'} isQuestionView={isQuestionView} activeSection={section} hearing={hearing} mainSection={mainSection} sectionGroups={sectionGroups}/>
+          <Col md={8} lg={9}>
+            <div className="section-browser">
+              <Link className="to-hearing" to={getHearingURL(hearing)}>
+                <Icon name="angle-double-left"/>&nbsp;<FormattedMessage id="hearing"/>
+              </Link>
+              <LinkWrapper disabled={!sectionNav.prevPath} className={`previous ${sectionNav.prevPath ? '' : 'disabled'}`} to={sectionNav.prevPath || '#'}>
```
Rikuoja: In the previous and next links, the section type name is selected based on the current section type even if we have sections of different types.

Here, sections 1-5 are called "Kysymys" and the last section is called "Hakulomake". Update the link texts depending on the section type.

Also, when we have several section types, only the first section type is listed in the sidebar. This should be fixed, though, when we display all sections in the sidebar regardless of where we are.
image

assets/less/hel/question-page.less
```diff
@@ -0,0 +1,35 @@
+.section-container {
+  .section-browser {
```
Rikuoja: image

src/views/Hearing/Sidebar.js
```diff
@@ -102,17 +104,29 @@ class Sidebar extends React.Component {
...
+                        <div>
+                          {getAttr(section.title, activeLanguage)}
+                          {'     '}
+                          <Icon name="comment-o"/>{section.n_comments}
+                        </div>
+                      </ListGroupItem>
+                    </Link>
+                  ))}
+                  {!isQuestionView && sectionGroups.map((sectionGroup) => (
+                    <ListGroupItem className={currentlyViewed === '#hearing-sectiongroup' && 'active'} href={"#hearing-sectiongroup-part"} key={sectionGroup.type}>
```
Rikuoja: Here, the section group is highlighted when hovering, but does not act as link when clicked.

src/components/SortableCommentList.js
```diff
@@ -140,7 +145,7 @@ class SortableCommentList extends Component {
 
         {canComment && <CommentForm
           hearingId={hearingId}
-          onPostComment={this.onPostHearingComment.bind(this)}
+          onPostComment={!onPostComment ? this.this.onPostHearingComment.bind(this) : onPostComment}
```
Rikuoja: For some reason, in the subsections, the comment form does not allow selecting a nickname at all. In the main hearing, the nickname box is shown. Is `canSetNickname` set right?

src/views/Hearing/Sidebar.js
jonnetanninen: Here is architectural challenge as these two use different "container" or "view" -component.
This is basically new feature and would take quite a lot of refactoring / writing new code.

Rikuoja: Hm, Ville had already added the subsections in the contents list? What happened to that?

jonnetanninen:
Suggested changes added excluding the on in 3rd comment

src/views/Hearing/Sidebar.js
Rikuoja: There is already a branch containing the subsection links in the sidebar, as intended and specified in the wireframes: https://github.com/City-of-Helsinki/kerrokantasi-ui/pull/282

Apparently, Ville closed the branch with the intention of integrating this one, and forgot to introduce the code into this branch. However, this feature is needed.

Rikuoja:
Hm, currently the sidebar doesn't scroll at all.

The idea is that the sidebar would be fixed to the page and scroll with the page. I believe that is the best we can currently attain.

PR #308, opened 2017-02-15
Eslint upgrade and slight cleanup

akx:
This upgrades Eslint and cleans up some of the complaints.

(Not all of them, because just looking at some of the expressions made me sad.)


Rikuoja:
I feel your sadness. BTW, yarn.lock seems sad too.

akx:
The lockfile has been unsaddened.

PR #311, opened 2017-03-09
Switch from less to sass

terotic:
I hope this does not break anyone's workflow. We use sass for common Helsinki themes, so makes sense to switch.


terotic:
After uneducated debugging and soul-searching I find that the test does not fail due to changing to sass. The same fail also took place with the previous commit and less-files. Maybe jest needs some configuring? 
https://travis-ci.org/City-of-Helsinki/kerrokantasi-ui/builds/202236526

PR #313, opened 2017-03-27
Display username as author when writing comment logged in

### LINKED ISSUE #263: Display user name as author name when writing a comment logged in, allow logged in user to change author name to a nickname that will be saved for future use

vaaralav:
Fixes #263

### DIFF
--- a/src/components/Section.js
+++ b/src/components/Section.js
@@ -136,7 +136,7 @@ export default class Section extends React.Component {
             data={section.plugin_data}
             onPostComment={this.onPostComment.bind(this)}
             pluginPurpose="postComments"
-            canSetNickname={user === null}
+            defaultNickname={user && user.displayName}
           />
         );
       case "map-questionnaire":
@@ -147,7 +147,7 @@ export default class Section extends React.Component {
             onPostVote={this.onPostVote.bind(this)}
             comments={comments}
             pluginPurpose="postComments"
-            canSetNickname={user === null}
+            defaultNickname={user && user.displayName}
             displayCommentBox={false}
             pluginSource={section.plugin_iframe_url}
           />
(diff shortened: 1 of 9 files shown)


src/components/Section.js
```diff
@@ -147,7 +147,7 @@ export default class Section extends React.Component {
             onPostVote={this.onPostVote.bind(this)}
             comments={comments}
             pluginPurpose="postComments"
-            canSetNickname={user === null}
+            defaultNickname={user && user.displayName}
```
Rikuoja: Here, the user default nickname should actually be obtained from the backend. I'm adding a `nickname` field to the API users endpoint.

Therefore, here we should change the `/me` endpoint to contain an additional user field `nickname` that is populated from the API `nickname` field, if the API has such a field. Then, we could set the nickname to be the one stored in the API.

The `user.displayName` should remain the user's real name, visible only to the user himself but not posted as nickname if a nickname already exists.

Rikuoja: The API already stores the posted nickname for future use by the same user.

src/utils/user.js
```diff
@@ -0,0 +1,20 @@
...
+  return user && user.nickname ? user.nickname : '';
+};
+
+/**
+ * [getAuthorName description]
+ * @param  {Object} user [description]
+ * @return {String}      [description]
+ */
+export const getAuthorName = (user) => {
+  return get(user, 'nickname') || get(user, 'displayName');
```
Rikuoja: The problem is, the user object here is obtained from the node.js frontend backend helsinki-passport app, which does not contain the nickname field.

If we add a nickname field, this will be obtained directly from the backend API. Is it possible to

a) get the author nickname and other properties directly from :https://api.hel.fi/kerrokantasi-test/v1/users/ endpoint or

b) plug the extra data from the API somehow into the passport user object?

jonnetanninen: If I'm understanding this correctly, both a & b should be possible. Where does one get the nickname at this point?

Rikuoja: Ah. My bad. Just added the nickname field to the backend master branch, so it should be available in the `users` endpoint `nickname` field.

PR #323, opened 2017-05-24
Support multiple languages in admin UI

### LINKED ISSUE #303: Refactor admin UI to function with multiple language backend

vaaralav:
Fixes #303

Still missing

- Nicely formatted and located error messages in Hearing Editor
- Figure out behaviour for Hearing Editor when user is not logged in

### DIFF
--- a/src/components/admin/HearingFormStep4.js
+++ b/src/components/admin/HearingFormStep4.js
@@ -10,7 +10,8 @@ import Row from 'react-bootstrap/lib/Row';
 import DateTime from 'react-datetime/DateTime';
 
 import {getClosureSection} from '../../utils/hearing';
-import TextArea from '../forms/TextArea';
+import {initNewSection, SectionTypes} from '../../utils/section';
+import MultiLanguageTextField, {TextFieldTypes} from '../forms/MultiLanguageTextField';
 import {hearingShape} from '../../types';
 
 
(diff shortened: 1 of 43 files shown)


src/components/admin/HearingFormStep4.js
```diff
@@ -92,12 +93,14 @@ class HearingFormStep4 extends React.Component {
           </Col>
         </Row>
 
-        <TextArea
+        <MultiLanguageTextField
```
Rikuoja: The closureInfo section is displayed here.

Therefore, closureInfo should not be shown at all in Form step 2 under hearing sections.

src/reducers/legacy/hearingEditor.js
```diff
@@ -0,0 +1,141 @@
...
+    image.url = "";
+  }
+  return assign({}, state, {hearing});
+};
+
+
+const addSection = (state, {payload}) => {
+  const hearing = cloneDeep(state.hearing);
+  hearing.sections.push(payload.section);
+  return assign({}, state, {hearing});
```
Rikuoja: When adding a new Section, it should open up expanded, while collapsing the previous section that the user was working on.

Rikuoja:
Also, I believe we should refactor the AdminHearings and AllHearings components to actually be available from the same url, `/hearings/`.

Whenever the logged in user is admin user for an organization, we would display the additional Publication Queue and Draft tabs for the user in the `hearings` list. No need for a separate address for the admin, methinks.

PR #328, opened 2017-05-30
Allow editing section title in hearing manager

### LINKED ISSUE #279: Allow editing section Title in admin UI

vaaralav:
Fixes #279


src/selectors/hearingEditor/labels.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+import type {LabelState} from '../../types';
+
+
+export const getAll = (state: LabelState): Array<Object> =>
+  state.all.map(frontId => state.byId[frontId]);
```
Rikuoja: It seems if there are more than 50 labels, only the 50 first ones are fetched from the server. Hence, the selector will return `undefined` for lots of labels and opening the manager will fail.

PR #346, opened 2017-07-06
Remove separate admin route

### LINKED ISSUE #341: Combine /admin and /hearings routes
No need for separate `/admin` route. Let's use the same listing page for mortals and admins, but let admins filter hearings based on their status (draft / publishing queue / published)

vaaralav:
Fixes #341


Rikuoja:
Also, for some reason beyond my understanding, `this.refs.frame` in `MapQuestionnaire.js` is now undefined, which results in any hearing with a map questionnaire failing to render completely. What does this have to do with the PR here?

Probably this well result in code failing in some other places where `this.refs` is used as well.

src/components/plugins/legacy/mapdon-ksv.js
```diff
@@ -20,7 +21,7 @@ class MapdonKSVPlugin extends BaseCommentForm {
 
   render() {
     const canSetNickname = this.props.canSetNickname;
-    const buttonDisabled = this.submitting || (!this.state.commentText && !this.state.userDataChanged);
+    const buttonDisabled = this.submitting || (!this.state.commentText && !this.state.user.dataDataChanged);
```
Rikuoja: Typo here, careful with the search and replace  :P

PR #350, opened 2017-07-10
Add localized error messages

### LINKED ISSUE #344: Display error messages/notifications in the selected language instead of Finnish
E.g. Ääni vastaanotettu

Rikuoja:
Fixes #344

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -6,6 +6,7 @@ import Helmet from 'react-helmet';
 import Header from './components/Header';
 import Footer from './components/Footer';
 import {retrieveUserFromSession} from './actions';
+import config from './config';
 
 class App extends React.Component {
   getChildContext() {
@@ -17,6 +18,7 @@ class App extends React.Component {
 
   componentDidMount() {
     this.props.dispatch(retrieveUserFromSession());
+    config.activeLanguage = this.props.language; // for non react-intl localizations
   }
 
   render() {
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -1,6 +1,7 @@
 import {createAction} from 'redux-actions';
 import api from '../api';
-import {alert, notifySuccess, notifyError} from '../utils/notify';
+import {localizedAlert, localizedNotifySuccess, localizedNotifyError, notifyError} from '../utils/notify';
+import getMessage from '../utils/getMessage';
 import merge from 'lodash/merge';
 import parse from 'url-parse';
 
@@ -30,7 +31,7 @@ export function getResponseJSON(response) {
 export function requestErrorHandler(dispatch, fetchAction) {
   return (err) => {
     const callName = fetchAction ? fetchAction.type : "";
-    notifyError("API-kutsu " + callName + " epäonnistui: " + err + ". Kokeile pian uudelleen.");
+    notifyError(getMessage("APICallFailed").replace('{callName}', callName).replace('{err}', err));
   };
 }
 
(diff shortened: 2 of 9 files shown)


src/App.js
```diff
@@ -17,6 +17,7 @@ class App extends React.Component {
 
   componentDidMount() {
     this.props.dispatch(retrieveUserFromSession());
+    window.language = this.props.language; // for non react-intl localizations
```
akx: Why pollute `window`, when you have a perfectly cromulent [`config`](https://github.com/City-of-Helsinki/kerrokantasi-ui/blob/master/src/config.js) object?

src/utils/getMessage.js
```diff
@@ -0,0 +1,23 @@
...
+/* eslint no-param-reassign: ["error", { "props": false }]*/
+
+/**
+ * Uses getAttr to obtain translations for UI message strings outside react-intl.
+ * @param  {string} string - The identifier for the string to be localized
+ * @param  {string} lang - Language code for requested language
+ */
+
+const getMessage = (string, lang) => {
+  const translations = config.languages.reduce((object, langCode) => {
```
akx: Can you explain what this reduce does (in a code comment)? Additionally, it looks like something that shouldn't be done on every invocation of `getMessage`.

src/utils/getMessage.js
```diff
@@ -0,0 +1,23 @@
...
+
+/* eslint no-param-reassign: ["error", { "props": false }]*/
+
+/**
+ * Uses getAttr to obtain translations for UI message strings outside react-intl.
+ * @param  {string} string - The identifier for the string to be localized
+ * @param  {string} lang - Language code for requested language
+ */
+
+const getMessage = (string, lang) => {
```
akx: Perhaps `lang` should just default to `config.language` (when it's moved there from `window`)?

```javascript
const getMessage = (string, lang=config.language) => { ... }
```
(unlike in Python, default arguments in ES6 are evaluated on every invocation; [see this babeljs transpilation for proof](https://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=false&presets=es2015&targets=&browsers=&builtIns=false&debug=false&experimental=false&loose=true&spec=false&code_lz=MYewdgzgLgBA5gUygWQRCBDRMC8MAU0ATgJZhwA0MANhuTqGAGYlwB0t5ArlggJS4AfDADeMAL4BuIA))

src/actions/index.js
```diff
@@ -30,7 +31,9 @@ export function getResponseJSON(response) {
 export function requestErrorHandler(dispatch, fetchAction) {
   return (err) => {
     const callName = fetchAction ? fetchAction.type : "";
-    notifyError("API-kutsu " + callName + " epäonnistui: " + err + ". Kokeile pian uudelleen.");
+    notifyError(getMessage("APICall") + " " + callName + " " +
```
akx: I think this message oughta be one string:
```javascript
{
  apiCallFailed: 'API-kutsu {callName} epäonnistui: {err}. ....',
}
```
and the placeholders replaced here:
```javascript
notifyError(getMessage('apiCallFailed').replace('{callName}', callName).replace('{err}', err));
```

src/utils/getMessage.js
```diff
@@ -0,0 +1,24 @@
...
+/* eslint no-param-reassign: ["error", { "props": false }]*/
+
+/**
+ * Uses getAttr to obtain translations for UI message strings outside react-intl.
+ * @param  {string} string - The identifier for the string to be localized
+ * @param  {string} lang - Language code for requested language
+ */
+
+const getMessage = (string, lang = config.activeLanguage) => {
+  // construct the translations object that only contains existing translations for given string
```
akx: Okay, um, I don't see a reason for using `reduce` here. Remember that `getMessage()` can easily be a fast-path function (i.e. called often), so it should be fast.

Maybe something like this (with old-school `for` loops so we get more performance and the ability to exit early).

```javascript
const getMessage = (string, lang = config.activeLanguage) => {
  // The usual path – the language exists and the string is translated in it.
  if(languages[lang] && languages[lang][string]) return languages[lang][string];
  // Otherwise, return the text in the first available language...
  for(let i = 0; i < config.languages.length; i++) {
    const map = languages[config.languages[i]];
    if(map && map[string]) return map[string];
  }
  // Or if that fails, just return the original.
  return string;
}

PR #365, opened 2017-07-13
Use Bootstrap pager for section navigation

terotic:


src/components/SectionContainer.js
```diff
@@ -218,31 +216,36 @@ class SectionContainer extends React.Component {
...
-              >
-                <FormattedMessage id="next"/>&nbsp;
-                <span className="type-name">
-                  {getAttr(sectionNav.nextType || section.type_name_singular, language)}
-                </span>
-                &nbsp;
-              </LinkWrapper>
+              <Pager>
+                {!sectionNav.prevPath
+                  ? <Pager.Item previous href={getHearingURL(hearing)}>
```
Rikuoja: Instead of `href`, we must wrap all the items to the `react-router` `Link` component.

That way, the whole page is not reloaded when the link is followed; instead, react-router takes care that only the section is changed, and the header and other unchanging parts stay on the screen, which makes for a much faster and smoother experience.

terotic: Okelydokely. Must then just use Bootstrap classes instead of bootstrap-react components. I guess components do not take router into account.

Rikuoja: What happens if you just put the component inside the link, or do the pager items need href as an argument?

terotic: `<Pager.item>` expands as `<li><a>` -combo so better to use the vanilla-html tags directly  for now. Not a big hassle, just a bit uglier code.

PR #367, opened 2017-07-14
Fix/emphasize eye icon

### LINKED ISSUE #340: Emphasize the crossed eye icon in front of a nonpublic hearing (e.g. with an explanation), so the user will understand that the hearing is not publicly visible

jonraem:
Fixes #340.


src/i18n/fi.json
```diff
@@ -153,6 +153,10 @@
   "inLanguage-sv": "ruotsiksi",
   "inLanguage-en": "englanniksi",
   "hearingLanguages": "Kuulemisen tiedot syötetään",
+  "eyeTooltip": "Tätä kuulemista ei ole julkaistu eikä se ole julkisesti nähtävissä.",
+  "eyeTooltipOpens": " Tämä kuuleminen julkaistaan",
```
Rikuoja: Remove extra space at the start.

src/views/Hearing/Header.js
```diff
@@ -10,11 +10,13 @@ import getAttr from '../../utils/getAttr';
...
   render() {
-    const { hearing, activeLanguage, reportUrl } = this.props;
+    const { hearing, activeLanguage, reportUrl, eyeTooltip } = this.props;
     return (
       <div className="hearing-header well">
         <h1>
-          {!hearing.published ? <Icon name="eye-slash" /> : null}
+          {!hearing.published
+            ? <OverlayTrigger placement="bottom" overlay={eyeTooltip}><Icon name="eye-slash"/></OverlayTrigger>
+            : null}
```
Rikuoja: Can we get an `&nbsp;` between the eye and the title so it will look better?

src/components/Hearing.js
```diff
@@ -217,6 +219,21 @@ export class Hearing extends React.Component {
     this.setState({ showDeleteModal: false, commentToDelete: {} });
   }
 
+  getEyeTooltip() { // eslint-disable-line class-methods-use-this
+    const { formatMessage } = this.props.intl;
+    const openingTime = moment(this.props.hearing.open_at);
+    let text = <FormattedMessage id="eyeTooltip" />;
+    if (openingTime > moment()) {
```
Rikuoja: Here we should have `if (this.props.hearing.published && openingTime > moment())` so that drafts will get the correct eyeTooltip text.

Issue #338, opened 2017-07-05
When a hearing is in draft phase, add a button for publishing hearing. Current "Hyväksy luonnos" doesn't make any sense.


Rikuoja:

Rikuoja:
Or does it?

As the "Publish" button doesn't immediately publish the hearing but, rather, the hearing will go live at a specified time in the future, what is the best wording here?

Obviously, this links to issue #339 and the wording, use the terms here: Draft/Will be public at 12:00 tomorrow/Published.

Issue #337, opened 2017-07-05
When a hearing is published, add a button for making hearing non-public


Rikuoja:
Needed in emergencies where e.g. it contains data that should not have been published yet, or needs to be revised before it can be published again.

Rikuoja:
Next to "Sulje kuuleminen". Name the button "Peru julkaisu"?

jonraem:
Yes!

Rikuoja:
Fixed in #369

PR #369, opened 2017-07-14
Text fixes for hearing view

### LINKED ISSUE #339: When a hearing is published, tell the user when it will be visible online.
This means 1) the time the hearing opens, if it is in the future, or 2) telling the user the hearing is already available, if the opening time is currently in the past.

jonraem:
Fixes #339, #338, #337.


jonraem:
I'll also add the other text fixes here.

Issue #342, opened 2017-07-06
Fix hearing section browser components on mobile


Rikuoja:

Rikuoja:
![image](https://user-images.githubusercontent.com/9367712/27916217-1a96700e-6271-11e7-89a8-2caafcac0054.png)

Rikuoja:
Fixed in #375

terotic:
Actually fixed in #365 👯‍♂️ 
My bad!

PR #375, opened 2017-07-18
Use new visual style and logo  on footer

terotic:


terotic:
I would prefer to use `import` to display svg logo  
```
import HelsinkiLogoWhite from 'hel-bootstrap-3/src/assets/helsinki-logo-white.svg';
```
but I get an error. 
```
(function (exports, require, module, __filename, __dirname) { <svg id="helsinki-logo" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 495 [...]
SyntaxError: Unexpected token <
```
The loader doesn't seem to like svg-syntax (tried with different svgs).

terotic:
Fix #342

PR #377, opened 2017-07-18
Revise toolbar status styles

terotic:
Maybe an overkill with icons and variating alert styles, but let's try this.

### DIFF
--- a/src/components/admin/HearingToolbar.js
+++ b/src/components/admin/HearingToolbar.js
@@ -3,9 +3,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {injectIntl, FormattedMessage} from 'react-intl';
 
-import Button from 'react-bootstrap/lib/Button';
-import ButtonToolbar from 'react-bootstrap/lib/ButtonToolbar';
-import Col from 'react-bootstrap/lib/Col';
+import {Row, Col, Button, ButtonToolbar, Alert} from 'react-bootstrap';
 import Icon from '../../utils/Icon';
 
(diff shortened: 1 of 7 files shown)


src/components/admin/HearingToolbar.js
```diff
@@ -38,7 +36,7 @@ class HearingToolbar extends React.Component {
     ];
     const open = isOpen(hearing);
     if (open && hearing.published) {
-      statusLabel = <FormattedMessage id="published"/>;
+      statusLabel = (<Alert bsStyle="success"><Icon name="eye"/> <FormattedMessage id="published"/></Alert>);
```
Rikuoja: Don't know if the backgrounds are overkill, but I think the green background is too dark on my display. The black text "Julkaistu" doesn't stand out as it should. The lighter backgrounds work alright.

terotic: Ok. Should also file an issue about this in hel-bootstrap-3. Must use white text with success, like we use in danger.

Rikuoja: Ah. Should test the white text before filing an issue though.

Rikuoja: Oh, and did you style the (non-react) notify.js notifications too? The ones that pop up when you post things in the UI?

terotic: You mean the Alertify popups? I can have a look, I guess they are using the plugin's default style at the moment.

Rikuoja: Ah it was Alertify, those exactly.

Issue #355, opened 2017-07-11
Fix sidebar contents highlighting when scrolling


Rikuoja:

terotic:
Not sure yet why this no works. Will investigate further when checking the hearing-page contents layout.

Rikuoja:
Sorta fixed in #379

PR #379, opened 2017-07-19
Fix hearing section highlighting when page scrolled

terotic:


terotic:
Fix #355

terotic:
This would work better if each page section content could be wrapped in `<Waypoint>` element. But the child of Waypoint can not be (functional component)[https://github.com/brigade/react-waypoint/issues/159#issuecomment-284240789]

Now the highlight is not triggered correctly when scrolling backwards.

PR #384, opened 2017-07-20
Don't add empty elements. Fix styles

terotic:
* Do not show header or lead elements on closure info box if they have no content
* Fix closure info styles accordingly

### DIFF
--- a/src/components/Section.js
+++ b/src/components/Section.js
@@ -57,11 +57,14 @@ export class Section extends React.Component {
       return null;
     }
     if (section.type === "closure-info") {
-      return (
-        <h3 className="section-title">
-          {getAttr(section.title, language)}
-        </h3>
-      );
+      if (section.title) {
+        return (
+          <h3 className="section-title">
+            {getAttr(section.title, language)}
+          </h3>
+        );
+      }
+      return null;
     }
     const iconName = (collapsed ? "chevron-right" : "chevron-down");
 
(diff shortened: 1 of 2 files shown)


src/components/Section.js
```diff
@@ -169,7 +172,7 @@ export class Section extends React.Component {
       {titleDiv}
       <div className="section-content">
         {imageList}
-        {section.type !== "main" ?
+        {section.type !== "main" && section.abstract ?
```
Rikuoja: The problem here is that section.abstract is always present even if empty (an empty object evaluates as true due to ES6 silliness). `getAttr(section.abstract, language)` will return `undefined` if the object is empty, that should be good :D

akx: [`_.isEmpty`](https://lodash.com/docs/#isEmpty).

src/components/Section.js
```diff
@@ -57,11 +57,14 @@ export class Section extends React.Component {
       return null;
     }
     if (section.type === "closure-info") {
-      return (
-        <h3 className="section-title">
-          {getAttr(section.title, language)}
-        </h3>
-      );
+      if (section.title.isString) {
```
akx: I don't think these objects have `isString`...

src/components/Section.js
```diff
@@ -169,7 +172,7 @@ export class Section extends React.Component {
       {titleDiv}
       <div className="section-content">
         {imageList}
-        {section.type !== "main" ?
+        {section.type !== "main" && section.abstract.isEmpty ?
```
akx: ... nor do I think they have `isEmpty`.

You'll need to
```javascript
import isEmpty from 'lodash/isEmpty';
import isString from 'lodash/isString';
```
and
```javascript
if(isEmpty(...))
```

PR #398, opened 2017-08-10
Test newer node versions, switch to codecov

hukka:


vikoivun:
Looks good. Uses sane node versions and such.

PR #414, opened 2017-09-26
Upgrade to Webpack 3

akx:
Out with the old and busted (in fairness, Webpack 1 _is_ deprecated), in with the new hotness!

:exclamation: Hot reloading does break with this – looks like the required bits can't agree on a chunk path? I tried to wrestle with this for a while to not much avail... Could probably be fixed by switching to using webpack-dev-server proper for dev with proxy configured to deal with the API endpoints.


akx:
> However, the dev server seems to build and rebuild several times before being satisfied. At next boot, the dev server happily builds only once, but yields an empty page [...]

I think 772611b might fix that, or at least alleviate it.

Rikuoja:
Yeah that seems to help with stochastic reloading. Had no effect on the undefined `Symbol`, but it resolved itself randomly after some iterations of trying to build dev and production configurations a few times more. Fully expect the dev server problems to resurface as they have done before, so probably will be running mostly production server unless I really really need the debug features.

akx:
Yeah, I actually saw the same `Symbol` problem too, now. Really not quite sure what's with it – and googling doesn't help much, either! – but it might resolve itself eventually, somehow... 😕

Rikuoja:
I must say that in the previous version of webpack, the missing or undefined dependency was something completely different! So the requirements failure is, while not stochastic, at the very least version dependent.

akx:
@Rikuoja So is this in mergeable shape?

Rikuoja:
Will have to be rebased and tested on IE11. Will not merge before that.

akx:
@Rikuoja Rebased. Do you have IE11 test tooling on hand?

Rikuoja:
No, and neither do I have the time now :(

akx:
I downloaded an IE11 image off Modern.ie, and things seem to work, aside from some flickering when scrolling. As far as I'm aware, nothing in this PR has changed anything related to that, so I'd think it's a pre-existing problem.

See here: https://gfycat.com/gifs/detail/GloriousFeistyAntlion

Issue #390, opened 2017-07-21
Do not render comments section if commenting is disabled or there are no comments yet


Rikuoja:

jonnetanninen:
1. The commenting form is rendered inside comment section so I guess we still want to render it even though there are no comments yet?

2. Do we still want to render existing comments if the commenting has been disabled?

Rikuoja:
1) Yes, the form should be displayed if commenting is allowed, but not the empty headline below it.

2) Yes, the headline and comments should be displayed whenever there are comments, but not the form is commenting is not (or no longer) allowed.

I guess previous coding messed this up somewhat, that's why the headline appears even if no comments are present.

PR #428, opened 2017-10-18
Feature/full-screen-navigation-menu

jonnetanninen:
Solves #390 & #407 & #419 & #411


Rikuoja:
Rebase needed.

jonnetanninen:
Rebased. This rebase was quite complex, so would appreciate some additional testing from someone else than myself incase I missed something.

Rikuoja:
What I will do is actually cherry-pick only some of the commits and not merge the fullscreen menu yet. This is because our fullscreen plugin does not have several languages available at the moment, so we don't want to present the useless UI to the user at the present :/

Issue #381, opened 2017-07-20
Add sentry integration instead of alertifying the user of JS errors


Rikuoja:

jonnetanninen:
Do we still want to alertify some general message for the user if something fails?

Rikuoja:
Hmm, I would say a general Finnish message along the lines of "Something went wrong. We sent an error report to the developers."

jussiarpalahti:
Well, as my original #324 stated, there's already feedback form provided by Sentry which might be useful. Then again, it might not.

jonnetanninen:
Is something like "Something went wrong. Try again soon or refresh your browser." ok as this will usually solve the problem for the user?

Rikuoja:
That sounds sensible. Reloading fixes the most common reason of these problems, which is expiration of the oauth token.

Rikuoja:
Implemented in #432, absent actual integration to deployment.

PR #432, opened 2017-10-19
Feature/sentry integration #381

jonnetanninen:
Solves #381

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,7 @@
     "redux": "^3.0.4",
     "redux-actions": "0.13.0",
     "redux-logger": "^2.0.4",
+    "redux-raven-middleware": "^1.2.0",
     "redux-thunk": "^2.0.1",
     "sass-loader": "^4.1.1",
     "scrolltop": "0.0.1",
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "qs": "^6.5.1",
     "query-string": "^4.3.4",
     "ramda": "^0.24.1",
+    "raven-js": "^3.19.1",
     "react": "^15.6.1",
     "react-bootstrap": "^0.30.7",
     "react-datetime": "^2.8.10",
--- a/server/Html.js
+++ b/server/Html.js
@@ -14,6 +14,10 @@ export default class Html extends React.Component {
     return (
       <html lang="fi">
         <head>
+          <script
+          src="https://cdn.ravenjs.com/3.17.0/raven.min.js"
+          crossOrigin="anonymous"
+          />
           <meta charSet="utf-8"/>
           <meta httpEquiv="X-UA-Compatible" content="IE=edge"/>
           <meta content="width=device-width, initial-scale=1" name="viewport"/>
--- a/server/Html.js
+++ b/server/Html.js
@@ -14,10 +14,6 @@ export default class Html extends React.Component {
     return (
       <html lang="fi">
         <head>
-          <script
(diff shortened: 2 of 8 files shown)


jussiarpalahti:
Question: how this behaves if sentry is not configured? @jonnetanninen @Rikuoja

server/Html.js
```diff
@@ -14,6 +14,10 @@ export default class Html extends React.Component {
     return (
       <html lang="fi">
         <head>
+          <script
```
jussiarpalahti: Why not use raven's NPM package? If need be it can be included outside regular bundle from static assets with some webpack trickery.

jonnetanninen: Done

package.json
```diff
@@ -80,6 +80,7 @@
     "qs": "^6.5.1",
     "query-string": "^4.3.4",
     "ramda": "^0.24.1",
+    "raven-js": "^3.19.1",
```
akx: This is the only reference to `raven-js` in the entire PR... I'd imagine you'd have to at least 
```javascript
import Raven from 'raven-js';
window.Raven = Raven;
```
or something...

src/actions/index.js
```diff
@@ -28,10 +27,11 @@ export function getResponseJSON(response) {
   return response.json();
 }
 
-export function requestErrorHandler(dispatch, fetchAction) {
+export function requestErrorHandler() {
   return (err) => {
-    const callName = fetchAction ? fetchAction.type : "";
-    notifyError(getMessage("APICallFailed").replace('{callName}', callName).replace('{err}', err));
+    // eslint-disable-next-line
```
akx: Why are we disabling ESlint altogether here? If it's because Raven is not defined, [specify it as a global](https://eslint.org/docs/user-guide/configuring#specifying-globals) or better yet, import it.
```javascript
import Raven from 'raven-js';
```

src/index.js
```diff
@@ -5,10 +5,17 @@ import {render} from 'react-dom';
 import getRoot from './getRoot';
 import createStore from './createStore';
 import commonInit from './commonInit';
+import config from './config';
 
 require('es6-promise').polyfill();
 
 commonInit(function initReady() {
+  try {
+    // eslint-disable-next-line
```
akx: Same comment about ESLint disabling here.

src/actions/index.js
jonnetanninen: Fixed

jonnetanninen:
Good point @jussiarpalahti . As is:

1. If sentryDns isn't defined will give error "Raven is not defined" but won't completely break the app.

2. If sentryDns is invalid the whole app will break -> don't know how to prevent this from happening

src/actions/index.js
jussiarpalahti: Is the eslint directive redundant now? Or necessary? I got confused 😕

jussiarpalahti:
Ok, now the missing sentry setting should function well.

Also, commenting on PRs is not exactly intuitive for me...

src/actions/index.js
jonnetanninen: Yes it's redundant and was removed as Raven is imported and thus defined now.

jussiarpalahti: But isn't it still active for that following line (numbered 32 and 33 I think) on this change?
Or am I looking at this wrong..?

jonnetanninen:
Rebased

Issue #406, opened 2017-09-19
Only load 10 or 20 top hearings on the list page at first


Rikuoja:
Load more when scrolling to the bottom. Currently, all hearings are loaded in chunks of 50, which makes for long requests and waiting.

PR #435, opened 2017-10-24
Feature/hearing list pagination #406

jonnetanninen:
Solves #406

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -34,6 +34,22 @@ export function apiCall(state, endpoint, params, options = {}) {
   return fetch(url, options);
 }
 
+export function apiCallWithUrl(state, url, options = {}) {
+  if (typeof state !== "object") {
+    throw new Error("API calls require redux state for authentication");
+  }
+  const user = getUser(state);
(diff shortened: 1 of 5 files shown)


src/api.js
```diff
@@ -34,6 +34,22 @@ export function apiCall(state, endpoint, params, options = {}) {
   return fetch(url, options);
 }
 
+export function apiCallWithUrl(state, url, options = {}) {
```
Rikuoja: Why do we need an extra helper utility for this single purpose, duplicating the API call code here? `fetchMoreSectionComments` seems to be using `api.get`, why don't we use the same here?

jonnetanninen: Actually I should've duplicated the api.get as I see now that the additional boilerplate from the apiCall wasn't necessary to make a successful call.

The idea was that by duplicating some code we would achieve the situation where we can use the "next" -url from the response without first parsing the url and immediately creating it again, which kinda doesn't make sense to me as we end up writing more code and it leaves us more prone to error. Also, we could use the function in the future if we end up doing more pagination somewhere else... But it's fixed now :)

PR #437, opened 2017-10-24
Determine plugin url by identifier

Rikuoja:
Waiting for transition to npm configuration, not mergeable yet.


Rikuoja:
Npm configuration not forthcoming. The server currently has a plugin redirect to a permanent directory one level up, not rewritten during deployment. However, new deployments will not have the needed files so we must also have the submodule deployment option available.

Issue #418, opened 2017-10-04
Create basic smoke tests for the UI


Rikuoja:
Five tests needed, perhaps with mock API and data:

1) front page
2) hearing listing
3) hearing page
4) posting a comment
5) hearing subsection page

Rikuoja:
There is already issue #145 about fixing tests and discussion on Jest, but we cannot fix tests if we don't have any :)

akx:
Jest has been switched to, so that's a non-issue. :)

jonnetanninen:
Is there more specification to this what exactly should be tested from each view?

jussiarpalahti:
I don't know what Riku had in mind, but I'd take current kerrokantasi state (from the site) and make enzyme checks and Jest snapshots based on that.

Rikuoja:
Current production version does work. The master branch does not :/

Rikuoja:
Done.

PR #444, opened 2017-11-08
Feature/smoke tests

jonnetanninen:
Solves #418 . 

The philosophy is to add snapshot tests with static mock date for the most important components as a "smoke test". This approach was approved and suggested by our lead developers. Later there is possibility to test certain functionalities separately and apply unit & integration tests, but I don't think it should be done in this phase as this PR would grow super large.

At the moment some of the tests are little bit questionable as the components tested are huge and super complex. To allow better tests and make testing easier, extensive refactoring would be required.


Rikuoja:
Oh, rebase needed on dependencies though.

Rikuoja:
One test that is more difficult to implement would be trying out what happens when the user tries to post a new comment in comment list.

This test would try out if the pipe from hearing sections all the way to posting a comment is correct. However, the result would require a mock API to display the comment back :/

jonnetanninen:
Rebased

jussiarpalahti:
@Rikuoja Is there particular reason we're not running tests through travis?

Can't we do so after this gets merged?

akx:
@jussiarpalahti Um... tests _are_ being run in Travis? https://travis-ci.org/City-of-Helsinki/kerrokantasi-ui/builds/303494223

jussiarpalahti:
Achoo.. Travis report was hidden behind that green thing among other green things.. I blame github.

All is well then 😁

Rikuoja:
Alas! Rebase needed!

jussiarpalahti:
Hm. I wonder why this wasn't mergeable? Perhaps it's just the order of changes that looks like a problem for git?

Rikuoja:
Most likely the snyk dependency or something else I just added caused the problem? :D

jussiarpalahti:
Arg.. Didn't scroll far enough :/
Snyk and this PR both change last lines of package.json
Oh well, it should be relatively trivial to merge..

jussiarpalahti:
Hrm. Github doesn't allow rebasing conflict resolution commit. Or I can't understand that part of the UI either. Oh well..

PR #447, opened 2017-11-11
Emslimmen Webpack bundle

akx:
This PR is a first step towards a slightly more manageable Webpack bundle.

Looking at stats.json, we go from 2085 modules and 6 MiB to 1686 modules and 5 MiB in the main bundle (dev build).

There's still work to do...


akx:
I removed the admittedly slightly superfluous server stuff – can be done in another PR.

Merging :)

PR #472, opened 2017-11-23
Update some dependencies and clean up a bit

akx:


akx:
Feel free to unaccumulate them :)

akx:
Don't suppose this could be merged?

Rikuoja:
Really depends on whether the unused state and props are still unused in master, after a considerable amount of bug fixing in these components.

akx:
@Rikuoja Rebased and split the prop/state removals into separate commits. On the way I found a little bug/unsupported case in the `eslint-plugin-react` plugin – the commit here is 511f454 and the fix for the plugin is https://github.com/yannickcr/eslint-plugin-react/pull/1628 .

Issue #423, opened 2017-10-12
Replace Hearing sidebar with a horizontal component


Rikuoja:
The problem is that the Sidebar is not very visual and does not scale with a large amount of sections. Therefore, a horizontal scrollable component is needed instead.

![image](https://user-images.githubusercontent.com/9367712/31496988-1eb0a014-af66-11e7-9630-05cdd8802728.png)

The functionality could be implemented e.g. with https://github.com/akiran/react-slick in Center mode. This way, the selected section will be scrolled to the center, also when switching the section with the Next/Previous section buttons (don't know how easy this will be to implement i'm afraid :/).

Therefore, *only* the sections (including the hearing itself, i.e. the main section) should be rendered, no headlines as in the old sidebar.

Rikuoja:
The map from the sidebar should be rendered as a thumbnail left of the main section. When clicked, it opens a larger (modal) view of just the map.

jonnetanninen:
Am I understanding correctly that this component shouldn't scroll down with the view like the sidebar did, but instead be static at the top of the hearing view?

Rikuoja:
Yes. Because actually the section nav, in this case, should be moved *under* the text content (but before the opinions), so the user will always get to the next section by clicking the next/previous buttons there, after reading one section through.

Therefore, there is no need to see this component if the user has scrolled down.

Rikuoja:
So yeah, this includes moving the section navbar component too, and including this component to the main page so the user can move from the hearing main page to the first subsection by clicking a button, if subsections are present.

Rikuoja:
Done.

PR #478, opened 2017-11-28
Feature/horizontal sidebar

jonnetanninen:
Solves #423


Rikuoja:
Though, of course, we still need `sectionGroups` for the `SectionBrowser`. But the main section will no longer display `sectionGroups` directly.

jonnetanninen:
Do not merge this yet. I found some bugs that have popped up in unexpected places due to the interesting component structure of the software.

Rikuoja:
Also, I'm thinking perhaps we should show several sections in the carousel on mid-sized mobile devices? The desktop and phone experience is fine, but a tablet would display the carousel with kinda confusingly empty space:

image

Rikuoja:
Oh, and apparently also the carousel will have to check the current route when rendering :( Because currently, if switching the sections with the browser, the carousel remains unmoved so that the current section might not even be visible in the carousel.

jonnetanninen:
Added scaling the amount of navigation items by screen width and added anchor links for more consistent navigation between sections. Latter is very hard to achieve, at least for the bottom of the section as we cannot place the anchor inside the content and the content height varies from one section to another...

terotic:
Yes. Agree. Next/previous could jump to anchor point.

Rikuoja:
Oh. One more thing.

There is a full screen Hearing (that has the Kartta koko näytöllä -button present) open all winter at kerrokantasi.hel.fi/talvipyoraily2017 .

Therefore, we will also have to support the case when there is a link to a full screen map plugin, much like the old Sidebar had the Kartta koko näytöllä -button.

This means that if a Hearing has a Plugin, we should render an extra button to the right to the main section, that takes us to the full screen map. Similar to the extra button to the left that displays the area map.

Rikuoja:
The "Full screen map" button should have a suitable graphic that indicates that it will open the full screen map. Does @terotic have any idea what such a button would ideally look like? We don't really want to render the clickable map component there, though it does in theory support rendering in any size :)

Rikuoja:
There is already a button at the bottom of the main section, but we should definitely add a similar button to the carousel itself, too:

image

jonnetanninen:
How would it seem now?

Rikuoja:
What says @terotic ? Now the carousel and the links at the bottom seem to work consistently, and so does the Full screen button in the carousel.

Rikuoja:
However, two smaller issues that actually popped up in user testing yesterday.

1) There is no longer any need to display the section name type ("osa-alue"/"vaihtoehto"/"kilpailutyö"/whatnot) in the Edellinen/Seuraava buttons either. I guess showing this name comes from the old sidebar, which has the sections displayed and grouped by type.

As we no longer have section groups in sidebars, we can get rid of the whole section grouping and naming logic in my opinion. The backend might provide the name of the section (Osa-alue etc.), but we don't need to display it here anywhere anymore, as there is no header that shows the list in text.

So let's just change the buttons to "<Edellinen", "Seuraava>" in all views. I think this simplifies the code and we can get rid of some useless parts that managed the section groups and section names in sidebar and browser, if I'm not wrong?

2) There is still one functionality missing that was present in the old sidebar. That is, when the hearing has several languages, there should be the listing of hearing languages ("Kuuleminen suomeksi", "Hearing in English" etc.) in the Header, as the Sidebar that provided that no longer exists.

Therefore, let's dig the removed Sidebar code for `getLanguageChanger` (https://github.com/City-of-Helsinki/kerrokantasi-ui/blob/master/src/views/Hearing/Sidebar.js#L91 ) and put it in the Header (all languages in one line). My suggestion would be to put it in the bottom right of the Header, on the same line as the social links, as there is ample free space:

image

However, perhaps @terotic has a better idea on what is the ideal location for the Hearing language changer? The languages should be, I believe, separated the same way as in the UI language changer: "Kuuleminen suomeksi | Hearing in English"

jonnetanninen:
Now should be mergable

PR #484, opened 2017-12-04
Code splitting

hukka:


Rikuoja:
Looks like they are loaded depending on the route, but do not wait for mounting, since the editor is not mounted at all but shares the same route with the hearing.

akx:
Hmmmm, adding deferring to `react-code-splitting` seems rather simple. Let me just make a PR...

Rikuoja:
Now implemented with react lazy.

Issue #480, opened 2017-11-29
Add the defaultClosureInfo string as default to editor sulkeutumistiedote


Rikuoja:
Currently, the "Sulkeutumistiedote" field is empty when a new Hearing is edited, and an empty string is saved. This means the `defaultClosureInfo` is never used in a Hearing, and the user sees an empty yellow box when an edited Hearing is closed.

Add the `defaultClosureInfo` string from intl as the default string already present, so it is always present when creating a Hearing.

Issue #475, opened 2017-11-24
Add error message for user trying to upload an image that is too large


Rikuoja:
Check the image size when added to section. Now, the user is presented with "Tarkista kirjautumisen tiedot" with no way of knowing that the reason is the image.

jonnetanninen:
The error handling from the response is pretty awkward at the moment, would it be ok to hard code the size validations to the UI and pre-emptively notify the user if he tries to upload image that is too big?

Rikuoja:
Definitely. Much better to validate the image before the UI even tries to send it :)

Rikuoja:
Validating in the UI based on the API responses is another thing, would require some rewriting plus checking the backend errors for consistency.

Issue #459, opened 2017-11-16
Translate instruction texts to finnish in hearing editor map


petterill:

jonnetanninen:
What does this mean more specifically? At the moment whole hearing editor should be in Finnish.

Rikuoja:
Ah. This seems to refer to the react-leaflet map tooltips etc.

jonnetanninen:
Does this require action?

Rikuoja:
Depends on whether react-leaflet is easily localizable :)

Rikuoja:
Done

Issue #456, opened 2017-11-15
In a non-public Hearing that is visible to admin, the editor bar at the bottom disappears when going to subsection views.


Rikuoja:
When entering a subsection page, the editor bar is no longer visible. Then, going back to the Hearing main page, the editor bar has disappeared.

Only reloading the page will make the editor bar visible again.

Rikuoja:
Sadly, not fixed yet. This will be addressed in the related issue #487

Rikuoja:
When users create a hearing and look at the created sections, they are permanently removed from the admin UI mode and have to reload or log out and in again.

Could this issue be ameliorated without too much of a refactoring, just by checking the admin status of the user also when entering the hearing from the subsections? Currently, it is only checked when the hearing route is entered, which results in the bar disappearing and not appearing again :(

jonnetanninen:
This is actually not a bug but a feature and needs #487 to be solved before this. At the moment the hearing editor component lives inside "HearingView" and subsection are inside "QuestionView". Thus the hearing editor won't work inside QuestionView/Subsections.

jonnetanninen:
This has was solved with PR #544

PR #485, opened 2017-12-04
Feature/#459

jonnetanninen:
Solves #459 & #456 & #455 & #475 & #453 & #433 & #480 & #397


src/utils/getTranslatedTooltips.js
```diff
@@ -0,0 +1,235 @@
...
+        }
+      },
+      simpleshape: {
+        tooltip: {
+          end: 'Vapauta painike lopettaaksesi piirtäminen.'
+        }
+      }
+    }
+  },
+  edit: {
```
Rikuoja: Is it possible to also translate `Zoom in` and `Zoom out` somewhere in here?

jonnetanninen: Unfortunately this is limitation caused by react-leaflet and thus impossible :(

jonnetanninen: 1. Can't recreate this, did you run yarn again after changing to this branch?

2. Darn, I mistook this for another bug. This is impossible to accomplish at the moment, because subsection views are actually QuestionView and the editor lives in HearingView. This would be achieved if QuestionView and HearingView were to be refactored into a single view (as it should be), but this is huge refactoring and shouldn't be part of this PR.

Rikuoja: Yeah agreed, refactoring is a separate issue, I take it we must defer it to https://github.com/City-of-Helsinki/kerrokantasi-ui/issues/487

And yes, probably yarn was the issue here, because trying again, I cannot reproduce the error anymore :)

Issue #473, opened 2017-11-23
When unselecting a language in the editor, do not send data in that language to the backend


Rikuoja:
Currently, the editor may have data that is not visible to the user (because they have unselected a language) but is posted to the backend.

We don't want to destroy any data that the user has input even if they clear the language selection for that language. Therefore, the data should still be stored, if the user clicks the checkbox by accident and doesn't want to delete all the texts.

However, when posting the hearing, we should only post those languages that are selected in the form when clicking the save button, so no invisible fields are sent to the backend.

PR #488, opened 2017-12-14
Feature/#473

jonnetanninen:
Solve #473


jonnetanninen:
We missed few other fields too, but I'm hoping that all fields are filtered now :)

PR #489, opened 2017-12-14
Fix #263

### LINKED ISSUE #263: Display user name as author name when writing a comment logged in, allow logged in user to change author name to a nickname that will be saved for future use

jonnetanninen:
Solves #263 . Replaces old PR #313 .

### DIFF
--- a/src/components/BaseCommentForm.js
+++ b/src/components/BaseCommentForm.js
@@ -46,7 +46,7 @@ export class BaseCommentForm extends React.Component {
   submitComment() {
     const pluginComment = this.getPluginComment();
     let pluginData = this.getPluginData();
-    let nickname = (this.state.nickname === "" ? null : this.state.nickname);
+    let nickname = (this.state.nickname === "" ? this.props.nicknamePlaceholder : this.state.nickname);
     let commentText = (this.state.commentText === null ? '' : this.state.commentText);
     let geojson = null;
     let label = null;
@@ -127,6 +127,13 @@ export class BaseCommentForm extends React.Component {
   }
 
   render() {
+    if (this.state.collapsed) {
+      return (
+        <Button onClick={this.toggle.bind(this)} bsStyle="primary" bsSize="large" block>
+          <Icon name="comment"/> <FormattedMessage id="addComment"/>
+        </Button>
+      );
+    }
     if (!this.state.collapsed) {
       return (<div className="comment-form">
         <form>
@@ -175,7 +182,7 @@ export class BaseCommentForm extends React.Component {
           <FormGroup>
             <FormControl
               type="text"
-              placeholder={this.props.intl.formatMessage({id: "anonymous"})}
+              placeholder={this.props.nicknamePlaceholder}
               value={this.state.nickname}
               onChange={this.handleNicknameChange.bind(this)}
               maxLength={32}
@@ -198,8 +205,8 @@ export class BaseCommentForm extends React.Component {
           </div>
           <CommentDisclaimer/>
         </form>
-      </div>
-    );
+      </div>);
+    }
   }
 }
 
--- a/src/components/BaseCommentForm.js
+++ b/src/components/BaseCommentForm.js
@@ -134,79 +134,77 @@ export class BaseCommentForm extends React.Component {
         </Button>
       );
     }
-    if (!this.state.collapsed) {
-      return (<div className="comment-form">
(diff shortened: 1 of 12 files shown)


src/components/BaseCommentForm.js
```diff
@@ -13,7 +13,7 @@ import forEach from 'lodash/forEach';
 export class BaseCommentForm extends React.Component {
   constructor(props, context) {
     super(props, context);
-    this.state = {collapsed: true, commentText: "", nickname: "", imageTooBig: false, images: []};
+    this.state = {collapsed: true, commentText: "", nickname: props.defaultNickname || 'Mörö', imageTooBig: false, images: []};
```
Rikuoja: I'm afraid the default nickname in absence of props will just have to be empty :(

src/components/plugins/MapQuestionnaire.js
```diff
@@ -40,18 +39,16 @@ class MapQuestionnaire extends BaseCommentForm {
...
-            <FormControl
-              type="text"
-              placeholder={this.props.intl.formatMessage({id: "anonymous"})}
-              value={this.state.nickname}
-              onChange={this.handleNicknameChange.bind(this)}
-              maxLength={32}
-            />
-          </FormGroup>
-        ) : null}
+        <h3><FormattedMessage id="nickname"/></h3>
```
Rikuoja: I have happy tidings: The `commentBox` is actually not needed in `MapQuestionnaire.js` anymore. It's a leftover from the old legacy map questionnaires.

The only place where `MapQuestionnaire` is rendered is in `PluginContent.js`, which automatically sets `displayCommentBox` prop to `false`. Therefore, we could just get rid of the prop `displayCommentBox` and all `commentBox` functionality here.

src/components/plugins/legacy/mapdon-ksv.js
```diff
@@ -14,7 +14,7 @@ class MapdonKSVPlugin extends BaseCommentForm {
   constructor(props) {
     super(props);
     this.pluginInstanceId = "ksv" + (0 | (Math.random() * 10000000));  // eslint-disable-line no-bitwise
-    this.state = Object.assign(this.state, {userDataChanged: false});
+    this.state = Object.assign(this.state, {userDataChanged: false, nickname: this.props.defaultNickname || ''});
```
Rikuoja: We actually won't need to update the legacy plugins, because the code is not to be used again in future hearings, so there's no need for the nickname functionality there. As long as the plugin renders when commenting is not enabled.

Rikuoja:
Could it be that login cannot be done after rendering the Hearing? I assume the props should automatically updated when logging in, though, and the user name should propagate all the way to the input form?

src/components/BaseCommentForm.js
jonnetanninen: Oh, this is embarrassing leftover from testing :D Will be removed

jonnetanninen:
If you have posted earlier with a nickname it should be rendered there after logging in. At the moment your real name will not appear in the "Nimimerkki" box, that is by design. I guess it should still appear to your comment if nickname is not specified?

The functionality gets a bit weird if you leave the box empty and thus post the comment without the author_name field. It seems that in this case the nickname of the user will be used as the comments default author_name in the backend.

I suggest that if the author_name in the post is left out, the former nickname should be removed and users full name should be used as author_name or we should implement some user info handling to the UI which I think at this point is not what we want...

Rikuoja:
1) I believe we want the actual user name to be the default nickname visible in the box for user. That is because if they don't write a nickname, the user name is actually shown next to the comment. Showing the user name will communicate to the user that if they don't type in a nickname, their real name will be visible. This has caused some confusion in the past, as users did not realize their real login name is visible when they comment.

2) I agree that if the user empties the field, then the real name is used again as the nickname. We have no other way of generating a nickname, and we don't want to allow empty nicknames.

jonnetanninen:
Ok so this is the functionality I ended up using and recommend. I noticed that for example my user's both first & last name was an empty string, so using them as default might not be viable option and could cause some weird functionality.

Now the placeholder text at least matches the default nickname which can also be changedof course.

What do you think @Rikuoja ?

jonnetanninen:
Now it should work as intended. Could not test properly though, as my user does not have displayName at all and Tunnistamo wouldn't let me log in with other methods :(

PR #494, opened 2017-12-18
Subnav carousel styling

terotic:
- Make carousel its own section
- improve responsiveness
- include map in the carousel
- add indicator for current (Active) section. @jonnetanninen : this needs functionality. Now highlights the center slide, not the currently active view.
- use default image on imageless sections


jonnetanninen:
Seems legit

PR #514, opened 2018-01-10
Solve #469: info texts

jonnetanninen:
Solves #469


Rikuoja:
Looks like snapshots conflict :(

PR #517, opened 2018-01-12
Fix warnings about bad DOM in SortableCommentList

juhana:
Removes console warnings about invalid DOM as per https://github.com/yahoo/react-intl/issues/286#issuecomment-236348944

### DIFF
--- a/src/components/SortableCommentList.js
+++ b/src/components/SortableCommentList.js
@@ -195,15 +195,12 @@ export class SortableCommentList extends Component {
                           onChange={event => {
                             this.fetchComments(section.id, event.target.value);
                           }}
+                          value={get(sectionComments, 'ordering')}
                         >
                           {keys(ORDERING_CRITERIA).map(key =>
-                            <option
-                              key={key}
-                              value={ORDERING_CRITERIA[key]}
-                              selected={ORDERING_CRITERIA[key] === get(sectionComments, 'ordering')}
-                            >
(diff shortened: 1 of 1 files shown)


Rikuoja:
I'm not getting any console warnings, what are the steps to reproduce the problem?

juhana:
If you open any page with a select element, for example any hearing with comments, you get:

```
Warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>.
Warning: Only strings and numbers are supported as <option> children.
Warning: validateDOMNesting(...): <span> cannot appear as a child of <option>. See SortableCommentList > option > FormattedMessage > span.
Warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>.
```

The warnings are generated by warning.js, they don't show in production. The DOM is invalid because `FormattedMessage` wraps the text in `<span>` inside select options.

Rikuoja:
Ah, now that you mention it I have been wondering about them. Currently I cannot get the development mode to run locally, so didn't get those warnings. Looks good!

Rikuoja:
Ah damn. 

Snapshots fail and lint fails, so these will still need to be fixed :(

juhana:
Right, sorry, I don't know how to update the snapshots.

Rikuoja:
Ah @jonnetanninen will have to do that I believe.

PR #560, opened 2018-01-30
Fix #546

### LINKED ISSUE #546: Legacy plugin mapdon-ksv.js does not render
While the code is legacy and will not be used in the future as is, the map is still needed to display the old hearing data. Currently, the `mapdon-ksv.js` component crashes at `super().ComponentDidMount`, so most likely there is something missing in the props as it is a subclass of `BaseCommentForm`?

jonnetanninen:
Solves #546


jonnetanninen:
Oh, I missed that feature completely. Setting displayVisualization true seems to work though

jonnetanninen:
Solved 💪

Issue #429, opened 2017-10-18
Display correct language for hearing label


Rikuoja:
Currently, if a Label has several languages, for some reason the Swedish label text is displayed on the Finnish hearing page. Check and fix Label language selection.

jonnetanninen:
I think this is a mix of several issues:

1. Some labels have Swedish string value behind "Fi" -key

2. Selected labels have the url as the source of truth. If it is for example "?label=raitiovaunut", when changing the language the url won't change and thus won't display the translated label. I don't think it possible to navigate to new url according the translated labels in any reasonable way.

Rikuoja:
I'm thinking we should rework the label filtering logic by id, not name.

The name of the label shouldn't come from the URL, but by the label name in the current language.

Also, the hearing page displays labels in the wrong language. There, the name cannot come from the URL but the label itself; the hearing url does not display any of the labels associated in the hearing.

Rikuoja:
But yeah, there definitely can be strings in the wrong fields in the backend atm, but that can be fixed. Regardless, the UI should pick the desired language field when looking at a Hearing with Labels.

PR #576, opened 2018-02-02
Solve #429 correct language for labels

jonnetanninen:
Solves #429


Rikuoja:
image

jonnetanninen:
Fortunately problem was simple, couple of missing props :)

PR #591, opened 2018-02-09
Add npm script fetching plugins

vikoivun:
This adds a script and a NPM command for fetching plugins. This makes it possible to always fetch the latest plugins, without needing to update the kk-ui repository itself (I understand submodules lock the versions). It also allows fetching the plugins to an arbitrary path.

I also added some documentation to the README about the existence of plugins. That still needs some cleanup. Information is welcome.

The list of plugins is in the script still. Would that need to be fetched from somewhere else as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Kerrokantasi UI is the user interface powering kerrokantasi.hel.fi service. It
 is a full featured interface for both answering and creating questionnaires as
 supported by Kerrokantasi API.
 
-## Development installation
+## Installation
 
 ### Prerequisites
 
@@ -21,13 +21,14 @@ supported by Kerrokantasi API.
 is in TOML-format, which for our purposes is `Key=Value` format.
 
 When NODE_ENV=="production", only environment variables are used for
-configuration. The environment variables are named identically to the
-ones used in config_dev.toml. Do note that the variables are case
-insensitive, ie. `KeRRokanTasi_aPi_bASe` is a valid name. Go wild!
+configuration. This is done because we've had several painful accidents
+with leftover configuration files. The environment variables are named
+identically to the ones used in config_dev.toml. Do note that the variables
+are case insensitive, ie. `KeRRokanTasi_aPi_bASe` is a valid name. Go wild!
 
 In the repository root there is `config_dev.toml.example` which contains
 every setting and comments explaining their use. If you only want to give
-kerrokantasi-ui a test, all you need to do is:
+kerrokantasi-ui a test, all configuration you need to do is:
 `mv config_dev.toml.example to config_dev.toml`
 That will give you a partially working configuration for browsing test
 questionnaires in our test API.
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -23,9 +23,29 @@ Scripts
...
+
+Questionnaires can make use of plugins. A default set of plugins can be
+installed using `npm run fetch-plugins -- { target_directory }`. By default,
+the plugins are installed in `assets/plugins` and served out in
+`assets/plugins` URL prefix by kerrokantasi-ui.
+
+If you wish, you can also specify a target directory and serve out that directory
+as `assets/plugins` in the same URL space as kerrokantasi-ui. The urls for
+plugins are partially hardcoded in kerrokantasi-ui/src/components/PluginContent.js
+and partially somewhere else. Maybe returned by the API?
```
Rikuoja: Nope, the UI should be the single source of truth of plugin urls, matching the plugin identifier (provided by the API) with the URL. The URLs are no longer present in the API.

Rikuoja: The identifier-URL mapping will indeed need to be hardcoded somewhere in the UI, you may suggest your favorite location for them. It's a bit weird to have them partially in `kerrokantasi-ui/src/components/PluginContent.js` and partially `deploy/fetch_plugins.js`, isn't it?

Rikuoja: Or perhaps it's not weird.

vikoivun:
Oh. Däng it. I thought github told me that everything was still fine.

README.md
vikoivun: Plugin location are now in src/shared_config.json
"Shared", as in shared between plugin cloner and the UI itself

vikoivun:
> Trying to install from scratch, I am met with the problem that git submodules include those directories already, so they are present (but empty) when cloning the repo from Github.
> 
> Therefore, the script doesn't actually install anything when run, giving me
> 
> ```
> 
> assets/plugins/kerrokantasi-citybikes for https://github.com/jvanamo/kerrokantasi-citybikes already exists, skipping
> assets/plugins/kerrokantasi-budgeting for https://github.com/jvanamo/kerrokantasi-budgeting already exists, skipping
> assets/plugins/kerrokantasi-bikeracks for https://github.com/jvanamo/kerrokantasi-bikeracks already exists, skipping
> assets/plugins/kerrokantasi-winterbiking for https://github.com/City-of-Helsinki/kerrokantasi-winterbiking already exists, skipping
> ```
> instead.
> 
> This is not the ideal setup for a new deployment, so apparently we need to decide on whether we want to support git submodules in addition to npm, which I of course would like as it is the current way of deploying the plugins.

The script now tests, if the directory contains a git repository, instead of just testing if the directory exists

vikoivun:
> So I guess
> 
> 1. rebase is needed,
> 2. readme should be updated to tell where the plugin urls really are specified (for those plugins that are not hardcoded and are fetched), and
> 3. .gitmodules and the committed plugin directories should be removed, as they are not needed any more?

These are now done.

Issue #590, opened 2018-02-09
Fix: contact persons visible on all pages of hearing


petterill:
Should be only main section

Issue #579, opened 2018-02-05
Allow opening hearing images in fullscreen window


petterill:
Clicking the hearing image on hearing page should open it in a fullscreen window, if the image is wider than hearing page width. Maybe open in some kind of an overlay on top of hearing page. If image is huge and does not fit in fullscreen window, zooming in/out should be an option within this fullscreen window as well.

Issue #574, opened 2018-02-02
Horizontal navigation back button is always disabled on widest breakpoint


jonnetanninen:
This seems to be because of a bug/feature of react-slick. When centermode is active, slidesToShow of 5 will actually be 7. Should work by dropping the count to 4.

jonnetanninen:
Ok it seems that everything works fine when there is more slides, but exactly 5 slides does not work very well. Lowering the count to 4 will mess which slides have slick-active class...

Issue #568, opened 2018-01-30
Closureinfo displays html tags in plain text


jonnetanninen:
This is because before rework closure section used dangerouslySetInnerHtml and after rework it is displayed in plain text. Should be fixed by using inner html again to display old infos properly.

Rikuoja:
When was it actually reworked to only allow plain text? Because I think the specs have always been that the closure section needs to contain links, so it cannot be plain text :)

PR #592, opened 2018-02-12
Fix #535 & #542 & #541 & #409 & #529 & #549 & #590 & #579 & #568 & #506 & #557 & #574 & #594

### LINKED ISSUE #535: Make Text field look more like Abstract field in hearing editor
- Border around Text field (tekstisisältö) should be bold as in Abstract field (Lyhyt johdanto)
- the field should be taller by default, similar in size as the Abstract
- info text should be inside the field, as in Abstract

jonnetanninen:
Solves multiple small issues: #535 & #542 & #541 & #409 & #529 & #549 & #590 & #579 & #568 & #506 & #557 & #574 & #594


jonnetanninen:
Doned!

src/components/SortableCommentList.js
```diff
@@ -177,67 +178,65 @@ export class SortableCommentListComponent extends Component {
...
+                <Icon name="comment-o" />&nbsp;{get(sectionComments, 'count') ? sectionComments.count : '0'}
+              </div>
+            </h2>
+            {pluginContent}
+            {this.state.showLoader && (
+              <div className="sortable-comment-list__loader">
+                <LoadSpinner />
+              </div>
+            )}
+            {showCommentList && !this.state.showLoader &&
```
Rikuoja: I am thinking here we actually shouldn't require `!this.state.showLoader`.

This is because we would like the comment list header *and* the order filter selection to be visible during loading of comments, so the user can be seen what is being fetched and in what order.

src/components/SortableCommentList.js
```diff
@@ -177,67 +178,65 @@ export class SortableCommentListComponent extends Component {
...
-                  </form>
-                </div>}
+        {section.commenting !== 'none' &&
+        <div className="sortable-comment-list">
+          {commentForm}
+          <div>
+            <h2>
+              <FormattedMessage id="comments" />
+              <div className="commenticon">
+                <Icon name="comment-o" />&nbsp;{get(sectionComments, 'count') ? sectionComments.count : '0'}
```
Rikuoja: Here, we shouldn't use `sectionComments.count` but, rather, `section.n_comments`. The former is zero when fetching, but the latter always contains the corrent count. Very seldom do we need `sectionComments.count` anywhere.

PR #600, opened 2018-02-23
Make home page images visible again

terotic:
- Remove inline styles from LinkWithLang component
- Add default image option on Fullwidth hearing


Rikuoja:
Hotterfix was quick enough, so this hotfix is no longer needed :)

PR #614, opened 2018-03-02
Feature/#550 hide navbar in webview

dotanthanh:
Modification to hide the navbar when the page is rendered in another application's webview. Minor change to LinkWithLangComponent to preserve headless query param when navigating.


jonnetanninen:
Looking very good, couple of things:

1. I would rename the reducer from `isInWebView` => `headless` so the name would match the url parameter as in the future developers might not have the context like we do now and that might be confusing for them

2. I would write similar simple middleware as the language.js for changing the state, as redux state should only be changed via actions and the current solution might cause some weird issues in the future. This also allow allow that functionality of changing the state in the future if it is needed somewhere

3. `config_dev.toml.example` file was accidentally removed, that should be returned

jonnetanninen:
LGTM!

PR #620, opened 2018-03-21
Feature/#611 process visualization in admin ui

dotanthanh:
a visualization of processes in hearing Editor (admin UI)
Right now: 
- layout is not 100% similar to initial sketch (but only minor differences)
- default phase and default project have multiple language support, but the number of languages of that default phase/project will not be updated if there is modification to languages selected in HearingFormStep1 after that.
- otherwise pretty good (bugs are hiding away from me really well)


jonnetanninen:
Issues I've found so far:

1. Creating new hearing crashes the app due to 
```
Uncaught (in promise) TypeError: Cannot read property 'id' of undefined
    at HearingFormStep5.render (HearingFormStep5.js?ba8d:87)
    at eval (ReactCompositeComponent.js?063f:793)
    at measureLifeCyclePerf (ReactCompositeComponent.js?063f:73)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (ReactCompositeComponent.js?063f:792)
    at ReactCompositeComponentWrapper._renderValidatedComponent (ReactCompositeComponent.js?063f:819)
    at ReactCompositeComponentWrapper.performInitialMount (ReactCompositeComponent.js?063f:359)
    at ReactCompositeComponentWrapper.mountComponent (ReactCompositeComponent.js?063f:255)
    at Object.mountComponent (ReactReconciler.js?c56c:43)
    at ReactCompositeComponentWrapper.performInitialMount (ReactCompositeComponent.js?063f:368)
    at ReactCompositeComponentWrapper.mountComponent (ReactCompositeComponent.js?063f:255)
```

2. Adding language for the hearing does not add input field for existing phases and title, only for phases added after adding new language

dotanthanh:
Ok so I updated with some new commits. Right now the logic is like this:
- project support languages that are specified in the hearing, and the project itself's languages. Unset language in hearing will remove empty language field in phases and project. For example, if the hearing specified "en" and "fi", and 1 phase's title has "fi" and "sv" non-empty, then that phase's title wil be available in all three languages.
- non-empty project's fields, and fields from phase that has hearings will not be removed by languages setting.

src/i18n/fi.json
```diff
@@ -195,6 +196,13 @@
   "loginToComment": "Kirjaudu sisään kommentoidaksesi.",
   "loginToVoteComment": "Kirjaudu sisään äänestääksesi kommenttia.",
   "addGeojson": "Lisää geojson tiedostona",
+  "addProcess": "Lisää vaiha",
```
jonnetanninen: Here's a small typo, should be "Lisää vaihe"

src/middleware/hearingEditor.js
```diff
@@ -88,10 +88,31 @@ export const replaceWithDefaultProject = store => next => action => {
...
+  if (action.type === EditorActions.INIT_NEW_HEARING) {
+    const {entities} = action.payload;
+    const hearingId = head(Object.keys(entities.hearing));
+    const hearing = entities.hearing[hearingId];
+    if (isEmpty(hearing.project)) {
+      return next(updeep({
+        payload: {entities: {
+          hearing: {
+            [hearingId]: {
+              project: {id: '', title: '', phases: []}
```
jonnetanninen: Should this be `project: {id: '', title: {fi: ''}, phases: []}` instead?

src/i18n/fi.json
dotanthanh: fixed. Now all phases' titles and project names are mandatory. Other minor modifications as mentioned above are added.

dotanthanh:
Add some fixes for the case when no language is selected, the phase will disappear:
- now there will be at least 1 project's name and 1 phase's title (if there is a phase), even if no language is selected.
- project name that is not empty will not be removed by select/unselect language, except for default language (we dont want to modify existing data about project).
- request loop is fixed also

PR #637, opened 2018-03-29
Make hearing cards visually consistent on front page

terotic:


vikoivun:
Verified to work using local install.

vikoivun:
Though I must confess, that I'm very confused with this "meta visibility"-thing. Like, partially visible?

PR #638, opened 2018-03-29
Synchronize production with master

vikoivun:


vikoivun:
Merged without review due to lack of time for reviews.

PR #639, opened 2018-04-06
Feature/#532 editable contact person

### LINKED ISSUE #532: Make editing existing contact persons possible
Editing previously created tags and contact persons should be editable. Clicking the name of the tag or contact person should bring up an editor window similar to the window where the tag/person was created

dotanthanh:
Resolved #532.


src/i18n/fi.json
```diff
@@ -159,6 +159,8 @@
   "eyeTooltipOpensHours": "tunnin kuluttua. Se ei ole vielä julkisesti nähtävissä.",
   "eyeTooltipOpensDays": "päivän kuluttua. Se ei ole vielä julkisesti nähtävissä.",
   "createContact": "Luo uusi yhteyshenkilö",
+  "editContact": "Muokkaa yhteystietoa",
+  "save": "Tallentaa",
```
jonnetanninen: Should be: "Tallenna"

src/actions/hearingEditor.js
```diff
@@ -157,6 +158,28 @@ export function addContact(contact, selectedContacts) {
...
+export function saveContact(contact) {
+  return (dispatch, getState) => {
+    const url = `/v1/contact_person/${contact.id}/`;
+    const contactInfo = omit(contact, ['id']);
+    return api
+      .put(getState(), url, contactInfo)
+      .then(checkResponseStatus)
+      .then(response => {
+        if (response.status === 400) {
+          notifyError('Tarkista yhteyshenkilön tiedot.');
```
jonnetanninen: I think this should be something like "Sinulla ei ole oikeutta muokata yhteyshenkilöä." as this error comes when trying to edit person from a organization I don't have edit rights for

dotanthanh: Should we use the error message from the response to display instead ?

src/i18n/fi.json
dotanthanh: fixed.

src/actions/hearingEditor.js
dotanthanh: fixed.

PR #659, opened 2019-01-23
New Dockerfile

halarotu:


vikoivun:
Dockerfile rewritten couple of times after this.

Issue #662, opened 2019-05-13
Update react and other main libraries to up-to-date versions


Rikuoja:
The code base has not been touched for a year or so. Therefore, the first task should be to

1) chart which dependencies are particularly outdated and
2) update those to reasonably recent versions.

PR #668, opened 2019-05-17
Updating packages

farrukh8925:
#662


farrukh8925:
it appears so :) otherwise the environment would even start if the dependency was used

PR #669, opened 2019-05-20
663 miniature map

### LINKED ISSUE #663: The miniature map in a hearing does not work
Currently, the miniature map in a hearing that should display the geometry on the map does not work. The default area shown is not the hearing area, and scrolling is exceedingly buggy.

Do note that if section navigation is revamped, the location of the map on the page might change. Therefore, if the bug is only due to the component where the map is embedded (react-slick), first try to fix the map so it works outside the scrollable component.

farrukh8925:
# Fixes #663 

Screenshot 2019-05-20 at 13 51 00


farrukh8925:
@Rikuoja the problem is that the react-leaflet requires fixed dimensions of width and height to load properly. I made partially dynamic configurations for it. Doesnt give me troubles. Did you try local pull ?

Rikuoja:
@farrukh8925 Yeah my branch is locally up to date. Doesn't seem to load at first, then loads at some point quite randomly, and breaks again when I navigate back to the page. Really weird that sometimes it works but mostly not.

Rikuoja:
Yeah it seems to start working when I e.g. resize the page!

farrukh8925:
Yep, thats the react-leaflet issue. It requires static width and height. Which i gave it :(

farrukh8925:
@Rikuoja pushed latest changes, can you give this a shot :)

Rikuoja:
Wow yeah, now it works!

Rikuoja:
Apparently just some lint issues still.

PR #678, opened 2019-06-03
666 file upload

### LINKED ISSUE #666: Allow the user to upload files from the editor
Once https://github.com/City-of-Helsinki/kerrokantasi-ui/issues/665 is implemented, allow uploading files in the hearing editor. 

Files need to be POSTed in the `/v1/file` endpoint as multipart messages, with the correct section id.

farrukh8925:
# Fixes
#666 

File upload operation, along with editing name of the file (on blur), and changing order, as well as deleting the file/


farrukh8925:
@Rikuoja there are two alternate approaches i thought of:
1 - When we upload a file, with or without a sectionid, we get response object with file information, we take that information and add it to POST request (creating new hearing), when we save it, it will use the BE will use the files array from payload and assign appropriate files. (BE Implementation required)

2 - We hide the upload file section while creating a new hearing.

Rikuoja:
Let's use option 1) now that the backend should support it!

Issue #676, opened 2019-05-24
Highlight comments by admins with yellow heading


katjahenttonen:
When a KerroKantasi admin user leaves comments (with a slightly different comment form defined in #675), the comments should be highlighted with a yellow heading.  As shown in the attached picture, the heading contains a fixed text "Helsingin kaupungin työntekijä:",  the name of the employee,  the date of the comment and the unit of the employee.  
![kaupunginvastaus](https://user-images.githubusercontent.com/39044068/58422941-89671b00-809c-11e9-8ce7-59377a8da912.png) Please ignore wrong font in "Kaupungin vastaus".

katjahenttonen:
The design has slightly changed here, please see the new pictures enclosed. 
![city-employee-comment-to-hearing](https://user-images.githubusercontent.com/39044068/58804388-6bf4fc80-861a-11e9-958b-0778d95600bf.png)
![nestedcomments-open-thread](https://user-images.githubusercontent.com/39044068/58804433-88913480-861a-11e9-87d3-ff34b9f0e7cc.png)

Rikuoja:
Here, we will display the yellow bar if comment `organization` is not null. Best just display the name of the organization as it is read from the comment `organization` field directly.

Rikuoja:
Now the `organization` in the comment JSON should display the name of the author organization, so you can just display it here.

farrukh8925:
@katjahenttonen there is are 2 minor issues. One, the color is not available in style guide, the closest is ``#ffe977`` named ``engel``
second, there will be some visible offset to the admin comment to the left to acomodate the border. in the snapshots you provided, perticularly the second one, the border is out of bounds of the parent div.

farrukh8925:
@Rikuoja i also noticed that when posting a new comment, the default behaviour is that the comment is stacked to top and list is re ordered ignoring the filters. That is too causing the jumping after one comment as a sub comment.

katjahenttonen:
@farrukh8925  i apologize for a late comment, but please you the same shade of yellow as in the UI specs, i.e. R: 255 G: 247 B: 175.

Rikuoja:
Solved in #682

PR #682, opened 2019-06-10
674 nested comments

### LINKED ISSUE #674: Allow user to answer comments (nested comments)
The user should be able to answer a specific comment, instead of answering the whole hearing.  This function is achieved by adding a text link called "Vastaa" (eng. answer, sw. svara) under each comment. After the user clicks "Vastaa", the new comment form should open below the comment, which the user is replying to.  Answers to a comment should appear nested below the original comment. The UI design is presented in the following picture: [nestedcomments2](https://user-images.githubusercontent.com/39044068/58322475-beb80280-7e29-11e9-90d2-de690072f29d.jpg)

farrukh8925:
# Fixes
#674 
#676


assets/sass/kerrokantasi/_commentlist.scss
```diff
@@ -1,3 +1,12 @@
+@keyframes highlightComment {
+    from {
+        background-color: $hel-engel;
```
juhakujala: There should be no Helsinki specific variables used in the `kerrokantasi/` styles. Helsinki specific styling should be added under the `hel/` folder. This variable can be changed into `$theme-primary-light` here and these styles can be overridden inside `hel/` styles folder if needed.

assets/sass/hel/variables.scss
```diff
@@ -4,6 +4,7 @@ $light-gray:    #ebedf1;
 $dark-gray:     #525a65;
 $medium-gray:   #abb2bd;
 $black:         #000000;
+$hover-blue:    #0072c6;
```
juhakujala: `#0072c6;` this value is the same as `$hel-coat` couldn't that be used here?

assets/sass/kerrokantasi/_commentlist.scss
```diff
@@ -96,6 +105,47 @@
...
+        font-weight: bold;
+        margin-top: $margin-default;
+        padding: $padding-default 0 0 0;
+        text-align: center;
+        &__wrapper {
+          border-bottom: 1px solid $black;
+          padding: 0 $padding-small;
+
+          &:hover {
+            border-bottom: 1px solid $hover-blue;
```
juhakujala: This `$hover-blue` value needs to be defined in the `assets/sass/kerrokantasi/variables.scss` file as well with the `!default` flag. Without it being assigned the build will crash with white label styles applied.

A good value for this would be `$hover-blue: $theme-info;`

farrukh8925:
> One thing I noticed: when entering a page, the comments are ordered by popularity.
> 
> Then, when posting an answer to a popular comment, all of a sudden the comments are reloaded so that all the comments are now ordered by date instead.
> 
> Because the popular comment was old, the newly posted comment is not seen, but instead it seems to reload the whole comment list and display the newest top level comments on top.
> 
> Probably this is caused by the line
> `if (parentComment && Array.isArray(parentComment.subComments) && parentComment.subComments.length > 0)`
> so that if we are posting the _first_ subcomment, the subcomments are not loaded, but it loads all the comments instead.
> 
> So we have to find another way to check that the user did post a subcomment (even though the section was _not_ open), and then open the section and only display the subcomments.

that is going to be a little tricky. But can be done. User would have to wait a bit though. But, i do get the jumping between is going to be annoying.

assets/sass/kerrokantasi/_commentlist.scss
```diff
@@ -96,6 +102,56 @@
...
+    }
+
+    &__edit-links {
+        display: flex;
+        flex-direction: row;
+        justify-content: flex-end;
+    }
+
+    &__admin {
+        border-left: 20px solid $hel-engel;
```
juhakujala: This value needs to be changed as well

assets/sass/kerrokantasi/_commentlist.scss
farrukh8925: replaced by theme-primary-light

assets/sass/hel/variables.scss
farrukh8925: removed, now using them-info instead

assets/sass/kerrokantasi/_commentlist.scss
farrukh8925: removed, theme-info now being used instead

assets/sass/kerrokantasi/_commentlist.scss
farrukh8925: removed

farrukh8925:
@Rikuoja the behaviour is now modified, even if the reply section is closed, we will fetch the replys for the parent component, and jump to newly created reply.

Rikuoja:
Damn, I merged the PR tidying up the scss styles first, so now this one conflicts again :(

farrukh8925:
> Damn, I merged the PR tidying up the scss styles first, so now this one conflicts again :(

No worries, pushed with fixes.

PR #687, opened 2019-06-12
Pluggable cities assets

frwickst:
So this was a fun one, lots of trial and error, in the beginning, to figure out the nicest way to do this. So what have I done?

## What does it do?
The code enables other cities to create their own assets and themes for kerrokantasi. It removes as much as possible of the Helsinki base assets from the codebase and instead moves all Helsinki related things to /cities/assets. Since I did not want to create any disruption for the Helsinki devs, the Helsinki themes/assets got to live on in the same repository. However, the main point would be that themes/assets would/could be separated into a separate project and then be activated by installing them and changing only a single config parameter, `city_config`.

To keep the application working as it did before, things will default to the Helsinki styles and assets and everything should work as it did before the change. However, it is now easy to load the whitelabel styles by settings `city_config = "whitelabel"` or to change to a different cities theme/assets, for instance `city_config = "kerrokantasi-ui-turku"` (https://github.com/City-of-Turku/kerrokantasi-ui-turku).

## How was it done?
Mainly by setting correct aliases in the WebPack. This allowed me to load assets from different places in the code base. This also allowed for cross-importing things from kerrokantasi-ui into a theme/assets package. When all of that was done, it was mainly figuring out where things should go and moving things around and pointing to the correct files in the code.

## Docs
Documentation regarding the creation of themes/asset packages can be found in the README so I won't bother explaining them here. 

## Notes
1. This does not remove all Helsinki related things from the main codebase yet. There are still some hardcoded strings that are left, and for instance, authentication is still very much Helsinki specific. This is mainly the first step into making the project more city agnostic so that others can use the same codebase and just be able to change the styling and strings that they want to.

2. As can be noted in the readme under "Installing city specific assets" and then "Development steps", there are still some caveats to this when creating external packages that would be nice to fix, but due to time constraints these were unfortunately not figured out.


Rikuoja:
I' m asking @vikoivun to comment on this one too, because he will be the one responsible for deployment anyway :) Everything should work as before, but better that he knows what is happening.

frwickst:
Added `city_config` to `config_dev.toml.example`.

cities/helsinki/assets/js/piwik.js
```diff
@@ -0,0 +1,11 @@
+var _paq = _paq || [];
+_paq.push(["setDomains", ["*.kerrokantasi.hel.fi"]]);
+_paq.push(['trackPageView']);
+_paq.push(['enableLinkTracking']);
+(function() {
+  var u="//analytics.hel.ninja/piwik/";
```
vikoivun: This should be changed to //analytics.hel.ninja/
and below should be piwik.php -> matomo.
Neither are blockers for this PR though, compatibility aliases are on place for the foreseeable future.

frwickst: The entire Piwik thing should be made optional as well, I did however not have time to do that right now. I will ask someone to look into it for a future PR.

Regarding the address, I have not changed the content of that file and it was like this before these changes, so this is not really an issue with this PR as far as I'm concerned 🙂

Rikuoja:
Looks like everybody is happy with this one, so I'm merging.

PR #696, opened 2019-06-17
Remove Helsinki specific strings from the code and make them translatable

juhakujala:


juhakujala:
Ps. the translation additions are failing the first three commits since the tests were not updated.

PR #709, opened 2019-06-20
Fix some of the image alt texts

### LINKED ISSUE #707: Screen readers should skip all the irrelevant images
Either provide an empty `alt=""` text or hide the images from screen readers otherwise.

juhakujala:
Add an empty alt tag for images in comments since it's impossible
to give the image an alt tag. This will make the screen readers skip
these images.

Define alt text for main hearing image: If there's no title or caption
provided, use an empty alt tag. If a caption is defined but no title, use
caption. If the title for image is defined, it will be used instead.

Also don't display the image caption element at all if no caption is
provided.

Fixes #707

### DIFF
--- a/src/components/Hearing/Section/SectionImage.js
+++ b/src/components/Hearing/Section/SectionImage.js
@@ -2,13 +2,14 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Lightbox from 'react-image-lightbox';
 
-export const SectionImageComponent = ({image, caption, title, showLightbox, openLightbox, closeLightbox}) => {
+export const SectionImageComponent = ({image, altText, caption, title, showLightbox, openLightbox, closeLightbox}) => {
+  const defineImageAlt = () => altText || caption || title || '';
+
   return (
-    <div className="section-image" key={image.url}>
+    <figure className="section-image" key={image.url}>
       <img
         className="img-responsive"
-        alt={title}
-        title={title}
+        alt={defineImageAlt()}
         src={image.url}
         onClick={openLightbox}
         onKeyPress={openLightbox}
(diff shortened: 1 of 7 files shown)


src/components/Hearing/Section/SectionImage.js
```diff
@@ -3,12 +3,22 @@ import PropTypes from 'prop-types';
...
 export const SectionImageComponent = ({image, caption, title, showLightbox, openLightbox, closeLightbox}) => {
+  const defineImageAlt = () => {
+    let imageAlt = "";
+    if (caption) {
+      imageAlt = caption;
+    }
+    if (title) {
+      imageAlt = title;
+    }
+    return imageAlt;
```
farrukh8925: Shouldn't the caption take priority over title ? In general image alt terms ? because imageAlt in this case will almost always be title.

anywho. Perhaps this block can be a lot shorter without variable assignment.

```
const defineImageAlt = () => caption || title;
```

Rikuoja: In most of the current data, there is no title so the image caption is displayed as alt text.

However, in general, I believe caption to be always more descriptive than the image title, so I agree with @farrukh8925 that we should use the caption (or even better, alt text, if we want to add backend support for that).

Rikuoja:
Anyway, I added the `alt_text` field to backend images, so it may be used if we wish to add the option to add alt text in the editor UI in the future.

Therefore, I suggest the order of precedence is `alt_text || caption || title`.

juhakujala:
Thanks for the comments! I fixed it to be `alt_text || caption || title || ''` since all of the three values can be undefined so we need the empty string as a backup.

Issue #553, opened 2018-01-23
Transition to using OpenID Connect and use silent renewal of user token


Rikuoja:
To ensure that the user will actually stay logged in while editing the Hearing.

Rikuoja:
This issue includes starting to use an up to date version of our SSO Tunnistamo. Therefore, an OIDC implementation is needed in place of the old OAuth that kerrokantasi-ui currently uses.

Rikuoja:
```"oidc-client"```

Rikuoja:
E.g. https://github.com/City-of-Helsinki/kaavapino-ui has a recent implementation, we would want it to work similarly here.

Rikuoja:
Ah, apparently https://github.com/city-of-helsinki/helpt-ui/blob/master/src/util/user-manager.js#L9 is a simple way to do it.

It uses `redux-oidc` as well, which performs silent renewal of the token in the background, and allows us to use the OIDC Implicit Flow, where everything happens in the client side and we need no server component in the UI. (We will retain the `server.js` though for server side rendering reasons.)

Rikuoja:
So this describes the renewal of the token with implicit flow: https://www.scottbrady91.com/OpenID-Connect/Silent-Refresh-Refreshing-Access-Tokens-when-using-the-Implicit-Flow

farrukh8925:
well, i've spent a little time going about the process. The problem is configuration files and how we read configurations in kerrokantasi. Since everything is in ./server/getSettings (which can only be read at the time of compilation), it becomes a challenge to access those variables outside the scope of ./server/. That being said we can "inject" using DefinePlugin. But lets see, im experimenting with it, since im not that familiar with client side authentication process.

farrukh8925:
BTW, this would mean no more `passport-helsinki`, right ?

Rikuoja:
Yes, I believe we can get rid of `passport-helsinki` if we get OIDC authentication working.

Jogha:
Is there any progress regarding using `redux-oidc`. For Kerrokantasi Turku case we are thinking on transitioning to that, and if there is already something done it would help us. Also is there any known problems that might arise when transitioning?

Rikuoja:
I haven't seen any progress on this front, perhaps @farrukh8925 can tell if there is a WIP branch or something somewhere on this topic that @Jogha could continue on?

Rikuoja:
Seems to be functioning now!

PR #710, opened 2019-06-26
Make authentication pluggable for different themes

Rippentrop:
This PR continues on the whitelabeling / de-helsinkifying of the base code, from the work started by Frank. This time the focus is specifically backend authentication handler.

Doing some investigation I found the `auth.js` *seems* to be entirely self-contained, so it felt like a good solution to just let every theme implement their own `auth.js`. As before, this should keep working normally for Helsinki without any exra configuration or other changes.

The code changes are documented more thoroughly in the commits and readme.


Rikuoja:
Okay, after literally taking a shower and having some thoughts there, this is what I suggest would be best for the future of kerrokantasi-UI:

1) fnish the transition to OIDC (issue https://github.com/City-of-Helsinki/kerrokantasi-ui/issues/553). I don't know if @farrukh8925 has some WIP branch you could continue on and finish?

2) implement a general non-city-specific configuration of the OIDC client, so that we can use any OIDC identity provider, such as Keycloak, Auth0, Tunnistamo or whatever software each city is using.

As I will probably be on holiday, @vikoivun can help you with configuring OIDC to work at least with the city of helsinki SSO tunnistamo, and probably any other OIDC identity provider that you wish to use in Turku.

PR #714, opened 2019-07-16
[WIP] Perform logout by taking user through tunnistamo logout page

vikoivun:
Makes logout actually do a logout at tunnistamo. This would not previously work, as tunnistamo does not allow fetch access to the logout endpoint.

Note that this is still hardcoded for tunnnistamo-test.

### DIFF
--- a/__tests__/login-logout.js
+++ b/__tests__/login-logout.js
@@ -17,7 +17,7 @@ describe('Login/logout', () => {
         username: 'mock.von.user',
         provider: 'helsinki',
       }),
-      "/logout": "OK"
+      "/logout/?next=/": "{\"sso-logout-url\": \"/\"}"
     });
     replace(fetcher);  // Replace the /me endpoint with a mock.
     const store = createTestStore();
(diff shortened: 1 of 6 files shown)


Rikuoja:
@vikoivun Is there still something unfinished in this PR? Apparently a timeout problem remains?

Rikuoja:
@vikoivun Is this still unfinished? I believe this should be implemented in all of our clients?

vikoivun:
> @vikoivun Is this still unfinished? I believe this should be implemented in all of our clients?

The action has moved to tunnistamo side. We've found a standardized (OIDC) logout-endpoint and that is in the process of being spruced (or perhaps pined) out.

Rikuoja:
Okay, so does kerrokantasi use that endpoint yet? What needs to be done?

vikoivun:
The poor page has gotten dropped to the ditch. We need to decide if it is really wanted, and if so, finish the implementantion on tunnistamo side.

PR #719, opened 2019-09-03
Hearing page refactor

juhakujala:
localhost_8086_suomenkielinen-kuuleminen_headless=false lang=fi
localhost_8086_suomenkielinen-kuuleminen_headless=false lang=fi (1)
localhost_8086_turkutesti_lang=fi


src/utils/hearing.js
```diff
@@ -47,6 +47,14 @@ export function acceptsComments(hearing) {
   return !hearing.closed && (new Date() < new Date(hearing.close_at));
 }
 
+/*
+* Returns true if any of the sections in a hearing has commenting enabled or if the hearing has a plugin
+* else false.
+ */
+export function hasCommentableSections(hearing, sections, user) {
```
Rikuoja: What is this function used for? Couldn't find it anywhere else.

src/i18n/en.json
```diff
@@ -75,6 +75,23 @@
   "lang-en": "English",
   "lang-fi": "Suomi",
   "lang-sv": "Svenska",
+  "hearingAvailable-en": "Hearing available",
+  "hearingAvailable-fi": "Kuuleminen saatavilla",
+  "hearingAvailable-sv": "Hörsel tillgängliga",
```
Rikuoja: Should be "Hörandet tillgängligt".

src/i18n/fi.json
```diff
@@ -121,6 +121,23 @@
   "lang-en": "English",
   "lang-fi": "Suomi",
   "lang-sv": "Svenska",
+  "hearingAvailable-en": "Hearing available",
+  "hearingAvailable-fi": "Kuuleminen saatavilla",
+  "hearingAvailable-sv": "Hörsel tillgängliga",
```
Rikuoja: Should be "Hörandet tillgängligt".

src/i18n/sv.json
```diff
@@ -25,6 +25,25 @@
   "lang-fi": "Suomi",
   "lang-en": "English",
   "lang-sv": "Svenska",
+  "attachments": "Bilagor",
+  "pinnedComment": "Fastnålad kommentar",
+  "hearingAvailable-en": "Hearing available",
+  "hearingAvailable-fi": "Kuuleminen saatavilla",
+  "hearingAvailable-sv": "Hörande tillgängliga",
```
Rikuoja: Should be "Hörandet tillgängligt".

src/utils/hearing.js
juhakujala: It's used here https://github.com/City-of-Helsinki/kerrokantasi-ui/pull/719/commits/cf715a28bab9e861791d21b7acb9464b97b16f80#diff-abd7d8933838e02a52669dd5b27b93cbR85

juhakujala:
@Rikuoja 
> One more thing to fix: the map plugin seems to now contain an extra (kerrokantasi-navbar) that should be removed. There should only be one navbar in fullscreen, and no navbar at all when not in fullscreen.

That navbar is inside the map plugin iframe which comes from another repo/project.

Rikuoja:
> @Rikuoja
> 
> > One more thing to fix: the map plugin seems to now contain an extra (kerrokantasi-navbar) that should be removed. There should only be one navbar in fullscreen, and no navbar at all when not in fullscreen.
> 
> That navbar is inside the map plugin iframe which comes from another repo/project.

Ah okay! I wonder why it didn't render before when the plugin was used?

PR #720, opened 2019-09-03
Add accessibility info/statement page

juhakujala:
This page is not displayed by default and it can be displayed by changing the config variable.


juhakujala:
Fixed the conflict.

The reason is that not all the cities might have the same accessibility issues because of their own branding which is affecting colors etc. Also currently there's no content written for it so this won't break anything at the time of merging.

PR #747, opened 2019-11-06
Implement Tunnistamo OpenId Connect authentication

Jogha:
This PR will implement OpenID Connect authentication flow, with some added Tunnistamo API token stuff to top that off.

Redux-oidc package is the base for this implementation, and is doing a lot of heavy lifting here. Tunnistamo uses separated API tokens, so fetching those is implemented together with redux-oidc.


src/reducers/api.js
```diff
@@ -0,0 +1,25 @@
...
+};
+
+const fetchApiToken = (state) => ({
+  ...state,
+  isFetching: true
+});
+
+const receiveApiToken = (state, {payload}) => {
+  if (payload) {
+    return {isFetching: false, jwt: payload['https://auth.turku.fi/kerrokantasi']};
```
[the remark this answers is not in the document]
Jogha: Yes, somehow this was left out of configurations... I will add it there.

src/views/Hearing/HearingContainer.js
```diff
@@ -21,6 +21,7 @@ const HearingEditor = lazy(() => import(/* webpackChunkName: "editor" */'../../c
 
 export class HearingContainerComponent extends React.Component {
   componentWillMount() {
+    console.error("HEARTING CONTAINER");
```
tituomin: Should probably be removed.

tituomin:
Great work, a few small comments above.

src/config.js
```diff
@@ -4,8 +4,13 @@ const config = {
   pluginMap,
   languages: ['fi', 'sv', 'en'],
   activeLanguage: 'fi',
+  publicUrl: (typeof window !== 'undefined' ? window.PUBLIC_URL : null) || 'http://localhost:8086/',
   apiBaseUrl: (typeof window !== 'undefined' ? window.API_BASE_URL : null) || 'http://localhost:8000/',
   uiConfig: typeof window !== 'undefined' ? window.UI_CONFIG : null,
+  openIdClientId: typeof window !== 'undefined' ? window.OPENID_CLIENT_ID : null,
```
Rikuoja: Here we expose the client ID to the client for anybody to see. Is the client id public data??

vikoivun: It is indeed. It is included in the “authorize” call made in users browser.

Rikuoja: Ah yeah. And the client id may indeed be a legible string (like in https://github.com/City-of-Helsinki/kukkuu-ui/blob/develop/.env#L5), as it does not need to be secret.

vikoivun:
I can pick up the documentation side.

PR #798, opened 2021-04-09
#786 Change keyboard focus and OK button in Write comment -form 

JoonasNislin:
Made changes so that the BaseCommentForm is autofocused when Write comment is clicked. Also localizedAlert focuses on the alert message instead of OK button when screen reader is in use.


Rikuoja (approved):
Works for me!

PR #838, opened 2021-12-08
Deleted comments

ranta:
- Style deleted comments slightly differently
- Hide interactable links from deleted comments

refs. KER-16
related: https://github.com/City-of-Helsinki/kerrokantasi/pull/362

### DIFF
--- a/src/components/Hearing/Comment/index.js
+++ b/src/components/Hearing/Comment/index.js
@@ -534,12 +534,14 @@ class Comment extends React.Component {
               {!isReplyEditorOpen && canReply && this.renderReplyLinks()}
             </div>
             <div className="hearing-comment-votes">
-              <Button className="btn-sm hearing-comment-vote-link" onClick={this.onVote.bind(this)}>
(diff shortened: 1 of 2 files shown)


src/components/Hearing/Comment/index.js
```diff
@@ -478,7 +478,7 @@ class Comment extends React.Component {
         <div className="hearing-comment__comment-wrapper">
           {this.renderCommentHeader(isAdminUser)}
           {!this.props.isReply && this.renderCommentAnswers()}
-          <div className="hearing-comment-body">
+          <div className={`hearing-comment-body ${data.deleted ? "hearing-comment-body-disabled" : ""}`}>
```
hienous, line 481: I would personally use the already imported `classnames` package for conditional classnames instead of doing it manually. Using classnames this line could look like 
`<div className={classnames('hearing-comment-body',{'body-disabled', data.deleted})}>`
which would render as `<div class="hearing-comment-body body-disabled">`

PR #896, opened 2022-11-03
OGP meta tag modificaitons

Jurppe:
Use `react-helmet` to manage OGP tags.

* remove hearing meta tag declarations in `Html.js`
* let views manage their own ogp tags
  * this provides multi language support and more dynamic labeling for hearings

### DIFF
--- a/server/Html.js
+++ b/server/Html.js
@@ -3,17 +3,15 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 export default class Html extends React.Component {
-  getMeta() {
-    const {heroImageURL, hearingData} = this.props;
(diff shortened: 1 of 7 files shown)


vikoivun (dismissed):
OK. Although you might consider having the default contain something like "Nimettömän kaupungin...". Then this would not be biased towards any city. PO's choice.

SanttuA:
Have you tested that setting OGP metadata works with React Helmet? Setting OGPs with javascript at later stages might not get read meaning that for example sharing a URL to a hearing would not use the correct metadata.

SanttuA (approved):

PR #920, opened 2023-02-08
Convert end date 00:00 to 23:59

voneiden:
Requested feature so that picking an end date defaults to end of day (23:59:59) instead of start of day (00:00:00). [react-datetime](https://github.com/arqex/react-datetime/#react-datetime) didn't seem to provide any good way of doing this. Therefore here's a proposed solution where we always convert an end datetime that isSame as start of day into end of day. An unfortunate side effect is that it is no longer possible to select "00:00" as the end time, the earliest possible time for the day is now "00:01". 

 Attempting to manually input "klo 00.00" will convert into "klo 23.59".


arpargo (approved):
Toimii ja tämä sivuvaikutus oli myös hyväksytty ja todettu tarkoituksenmukaiseksi joten ei ongelmia.

PR #927, opened 2023-03-06
KER-110

arpargo:
Kartalla saattoi näkyä sulkeutuneita kyselyitä.

Syy: Default käsittely kartan renderöinnissä ei käyttänyt Math.random() arvoa avaimena vaan geojsonia. Geojson ei kuitenkaan ole uniikki vaan saattoi esiintyä useasti -> aiheutti virheen ja nämä toimivat kartalla tämän jälkeen satunnaisesti.

### DIFF
--- a/src/components/OverviewMap.js
+++ b/src/components/OverviewMap.js
@@ -131,7 +131,7 @@ class OverviewMap extends React.Component {
           default:
           // TODO: Implement support for other geometries too (markers, square, circle)
             contents.push(
-              <GeoJSON data={geojson} key={JSON.stringify(geojson)}>{this.getPopupContent(hearing, geojson)}</GeoJSON>
+              <GeoJSON data={geojson} key={Math.random()}>{this.getPopupContent(hearing, geojson)}</GeoJSON>
             );
         }
         // contents.push(<GeoJSON key={id} data={geojson}>{content}</GeoJSON>);


src/components/OverviewMap.js
```diff
@@ -131,7 +131,7 @@ class OverviewMap extends React.Component {
           default:
           // TODO: Implement support for other geometries too (markers, square, circle)
             contents.push(
-              <GeoJSON data={geojson} key={JSON.stringify(geojson)}>{this.getPopupContent(hearing, geojson)}</GeoJSON>
+              <GeoJSON data={geojson} key={Math.random()}>{this.getPopupContent(hearing, geojson)}</GeoJSON>
```
mikkojamG, line 134: Math.random does not return a unique number, although changes of duplicates is very small. Could still mix this random value with something to make it unique, date string or or the hearing id upper in the scope 🤔

mikkojamG (changes requested):

src/components/OverviewMap.js
mikkojamG: Would most likely to fix this for other switch cases too

arpargo: Well this is now implemented in smarter way, it still has a small chance of having duplicate key because a single hearing can have multiple markers but complicating this further would be close to pointless as there are no dangers associated with the map rendering incorrectly once in a billion tries.

mikkojamG (approved):

mikkojamG (dismissed):

arpargo:
Should test the Map view of the Hearings listing extensively as it has been refactored to have less code duplication.

src/components/OverviewMap.js
```diff
@@ -146,53 +92,78 @@ class OverviewMap extends React.Component {
...
+        }
+          break;
+        case "Feature": {
+          /**
+           * Recursively get the Map element
+           */
+          return (this.getMapElement(geojson.geometry, hearing));
+        }
+        default:
+          // TODO: Implement support for other geometries too (markers, square, circle)
```
hienous, line 160: This doesnt work correctly for a hearing with multiple markers, did you test that this works correctly?  The refactored getMapElement only takes 1 param (hearing) and here we are passing 2 params to it. Here is an image of the geojson object of a hearing that has multiple markers(a FeatureCollection)
image

arpargo: Yeah, I was worried something like this could happen. It seemed to work on my local tho as there was one test case with multiple point type notations and those were also correctly limited to 10 on home page view.

arpargo: Right, realized I had not rebuilt the thing and so it just seemed to still work, implemented a proper fix now @hienous, can you verify the fix works?

src/components/OverviewMap.js
```diff
@@ -147,52 +93,77 @@ class OverviewMap extends React.Component {
...
+              mapElementArray.push(this.getMapElement(feature.geometry, hearing));
+            });
+          } else {
+            // if mapElementLimit is false -> display all elements found in FeatureCollection
+            geojson.features.forEach((feature) => {
+              mapElementArray.push(this.getMapElement(feature.geometry, hearing));
+            });
+            return mapElementArray; 
+          }
+        }
```
hienous, line 151: `return mapElementArray` is currently inside the `else` part, move `return mapElementArray` to happen after the if / else sections (between curly brackets on lines 150 / 151).

src/components/OverviewMap.js
```diff
@@ -67,74 +67,20 @@ class OverviewMap extends React.Component {
...
-            contents.push(
-              <GeoJSON data={geojson} key={JSON.stringify(geojson)}>{this.getPopupContent(hearing, geojson)}</GeoJSON>
-            );
+        const mapElement = this.getMapElement(geojson, hearing);
+        if (Array.isArray(mapElement) && mapElement.length > 0) {
+          mapElement.forEach(mapEl => {
+            contents.push(mapEl);
+          });
+        } else if (mapElement && !Array.isArray(mapElement)) {
+          contents.push(mapElement);
```
hienous, line 82: this section could be simplified further, as the else if tests seem redundant. 
image

arpargo: I felt like adding empty arrays into the contents array would be sort of bad practice even if it currently does work. This way the contents array always only consists of objects that we want to show on the map and we're not reliant on the map library to handle the exceptions.

src/components/OverviewMap.js
arpargo: It does detect it but doesn't highlight it well with my current configurations, have to change it 👍

hienous (approved):

PR #956, opened 2023-08-04
KER-226 | Add instructions for using local API and Tunnistamo

charn:
This PR adds instructions for using a local API and Tunnistamo instances for development with docker.


mikkojamG (approved):
Works 👍

PR #959, opened 2023-08-10
KER-210 HDS footer

mikkojamG:
# KER-210 HDS footer

- Replace footer with the latest HDS Footer component
- Fix HDS base styles on whitelabel theme
- Add `settings.json` for further customising city themes. In this case it would be using different koros type for whitelabel and Helsinki themes. **Note:** Koros calm seems to work little incorrectly atm, as it uses too much space at the top of the footer. A bug ticket has been created on HDS board.
- Webpack 5 polyfills

**Helsinki:**

Screenshot 2023-10-16 at 12 48 16

**Whitelabel:**

Screenshot 2023-10-16 at 12 47 32


src/components/Footer.jsx
```diff
@@ -1,111 +1,94 @@
...
+          target='_blank'
+          rel='noopener noreferrer'
+        />
+      </HDSFooter.Utilities>
+      <HDSFooter.Base
+        copyrightHolder={<FormattedMessage id='copyrightHolder' />}
+        copyrightText={<FormattedMessage id='copyrightText' />}
+        backToTopLabel={<FormattedMessage id='scrollToTop' />}
+        logo={
+          <Logo
```
arpargo, line 61: This needs to be added to settings.json aswell to support Turku not using Helsinki logo.

arpargo (changes requested):
Apart from the other comment, everything seems to work as it should and looks like it's customizable enough.

src/components/Footer.jsx
arpargo: Nevermind, it works as requested...

arpargo (approved):
LGTM

PR #963, opened 2023-08-18
WIP Feature: upgrade to node18 & webpack 5

arpargo:
# Node18 päivitys

## Node18 ja webpack 5 päivitys

-----------------------------------------------------------------------------------------------

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,8 +92,9 @@
     "prop-types": "^15.7.2",
     "qs": "^6.11.2",
     "query-string": "^7.0.0",
+    "querystring-es3": "^0.2.1",
     "raven-js": "^3.27.2",
-    "react": "^16.13.1",
+    "react": "^16.2.0",
     "react-anchor-link-smooth-scroll": "^1.0.12",
     "react-bootstrap": "0.33.1",
     "react-bootstrap-switch": "^15.5.3",
@@ -129,13 +130,14 @@
     "simple-git": "^3.3.0",
     "snyk": "^1.695.0",
     "style-loader": "^2.0.0",
-    "terser-webpack-plugin": "^1.4.5",
+    "terser-webpack-plugin": "^5.2.5",
     "updeep": "^1.2.0",
+    "url": "^0.11.1",
     "url-loader": "^1.1.2",
     "url-parse": "^1.5.9",
     "uuid": "^3.4.0",
-    "webpack": "^4.46.0",
-    "webpack-dev-middleware": "^4.1.0",
+    "webpack": "^5.54.0",
+    "webpack-dev-middleware": "^6.1.0",
     "webpack-hot-middleware": "^2.25.1",
     "webpack-merge": "^4.2.2"
   },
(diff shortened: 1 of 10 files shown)


src/selectors/projectLists.js
```diff
@@ -1,7 +1,9 @@
-// @flow
-import type {AppState} from '../types';
-
-export const getProjects = (state: AppState): [Object] => {
+/**
+ * 
+ * @param {AppState} state
```
mikkojamG, line 3: Should probably replace with `object` or something like that, since AppState is no longer imported.

src/selectors/user.js
```diff
@@ -1,11 +1,23 @@
-// @flow
-import type {AppState, User} from '../types';
-
-export const getUser = (state: AppState): User =>
+/**
+ * 
+ * @param {AppState} state
```
mikkojamG, line 3: Should probably replace with `object` or something like that, since AppState is no longer imported.

src/selectors/projectLists.js
arpargo: fixed

src/selectors/user.js
arpargo: fixed

package.json
```diff
@@ -124,36 +125,39 @@
...
     "sass": "^1.62.1",
-    "sass-loader": "^7.0.0",
+    "sass-loader": "^13.3.2",
     "scrolltop": "0.0.1",
     "simple-git": "^3.3.0",
     "snyk": "^1.695.0",
-    "style-loader": "^2.0.0",
-    "terser-webpack-plugin": "^1.4.5",
+    "style-loader": "^3.3.3",
+    "terser-webpack-plugin": "^5.2.5",
```
mikkojamG, line 133: Seems there is a problem with the terser plugin when building Docker image:
`ValidationError: Invalid options object. Terser Plugin has been initialized using an options object that does not match the API schema.
#0 17.61  - options has an unknown property 'sourceMap'. These properties are valid:
#0 17.61    object { test?, include?, exclude?, terserOptions?, extractComments?, parallel?, minify? }...`

package.json
```diff
@@ -29,19 +29,19 @@
...
     "@babel/register": "^7.0.0",
     "@draft-js-plugins/drag-n-drop": "^4.2.0",
     "@draft-js-plugins/editor": "^4.1.0",
     "@draft-js-plugins/focus": "^4.1.1",
     "@draft-js-plugins/image": "^4.1.1",
-    "@draft-js-plugins/resizeable": "^4.1.0",
+    "@draft-js-plugins/resizeable": "^5.0.4",
     "alertifyjs": "^1.13.1",
-    "autoprefixer": "^8.6.5",
+    "autoprefixer": "^10.4.15",
```
mikkojamG, line 40: Got this warning when building, seems like an easy fix:
```Replace Autoprefixer browsers option to Browserslist config.
  Use browserslist key in package.json or .browserslistrc file.

  Using browsers option can cause errors. Browserslist config can
  be used for Babel, Autoprefixer, postcss-normalize and other tools.

  If you really need to use option, rename it to overrideBrowserslist.```

mikkojamG (changes requested):
Runs nicely when running with Node. Remember to update `.nvmrc` and Dockerfile Redhat image. Had some issues when building Docker image, seems related to terser plugin.

package.json
arpargo: Fixed

mikkojamG (dismissed):
LGTM 👌

mikkojamG (approved):

PR #968, opened 2023-09-18
KER-252 bug BaseCommentForm extends

mikkojamG:
# KER-252 bug BaseCommentForm extends

`MapQuestionnaire`, `MapdonKSVPlugin` and `MapdonHKRPlugin` all extend BaseCommentForm, that previously was refactored into function component. These components are really similar and even used in same places, so removed `MapdonKSVPlugin` and `MapdonHKRPlugin` entirely and refactored `MapQuestionnaire` to support legacy features.


arpargo (commented):
It seems previous refactor of BaseCommentForm has broken some backwards compatibility (https://kerrokantasi.dev.hel.ninja/testikuuleminen) and this seemed to almost fix it but failed due to a different error: Uncaught Error: Function components cannot have string refs. We recommend using useRef() instead.

mikkojamG:
> It seems previous refactor of BaseCommentForm has broken some backwards compatibility (https://kerrokantasi.dev.hel.ninja/testikuuleminen) and this seemed to almost fix it but failed due to a different error: Uncaught Error: Function components cannot have string refs. We recommend using useRef() instead.

Fixed in commit [5b6db22](https://github.com/City-of-Helsinki/kerrokantasi-ui/pull/968/commits/5b6db22abd5828fb2f710513b037333d9d38a3a5)

Jurppe:
Neat. Seems to be working and the implementation looks clean. 👍

Jurppe (approved):

PR #969, opened 2023-09-18
Separate contact person fetching to different dispatch method

Jurppe:
Fetch contact person data when mounting `<HearingEditor />` component. The parent `<HearingContainer />` does not need the data from `/v1/contact_person/` endpoint.  Eg. fetch the data from endpoint `/v1/contact_person/` only when the user is admin or editor of the hearing.

* separate the `/v1/contact_person/` api call from the bundled dispatch method `fetchHearingEditorMetaData()`
* perform this dispatch in `<HearingEditor />` constructor


mikkojamG (approved):
Seems to work as intended 👍

PR #982, opened 2023-10-24
KER-270: HDS Button

arpargo:
Refactoring away from Bootstrap Button component. Long term plan to remove all Bootstrap from the service as upgrading it is not simple and most of the components can be found from HDS.


src/components/Hearings/CreateHearingButton.jsx
```diff
@@ -11,7 +12,7 @@ const CreateHearingButton = ({ to }) => (
   <div className='toolbar-bottom create-hearing-button-container'>
     <ButtonToolbar className='actions pull-right'>
       <Link to={to}>
-        <Button bsStyle='success'>
+        <Button theme="kerrokantasi" className="kerrokantasi-btn success">
```
mikkojamG: `theme` can be removed

src/components/admin/HearingFormStep5.jsx
```diff
@@ -110,7 +110,7 @@ class HearingFormStep5 extends React.Component {
         </div>
         {selectedProject && (
           <ButtonToolbar>
-            <Button className={errorStyle} onClick={this.addPhase} bsSize='small' bsStyle='default'>
+            <Button className={`${errorStyle  } kerrokantasi-btn`} onClick={this.addPhase} size='small' >
```
mikkojamG, line 113: Coud replace with `classNames([errorStyle, 'kerrokantasi-btn'])`

mikkojamG (changes requested):
Good work, this is a large operation so it must be hard to keep track of every button 👍 Had lint errors in `HearingsSearch.jsx`, probably due to rebase: `hearings-search__controls` missing closing div.

Should this be in color black?
Screenshot 2023-11-06 at 12 35 45

`Peru` and `Lähetä` could probably use a bit of space in between. Also should `Peru` have black border and color?
Screenshot 2023-11-06 at 12 37 04

mikkojamG (dismissed):

mikkojamG (approved):

PR #983, opened 2023-10-25
KER-277 Fix language from query string bug

timwessman:
Hotfix for a bug: [KER-277](https://helsinkisolutionoffice.atlassian.net/browse/KER-277)

So, getting language from query string didn't work. There was a bug on middleware that parses query string -> now fixed

[KER-277]: https://helsinkisolutionoffice.atlassian.net/browse/KER-277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #985, opened 2023-10-25
KER-277 fix language change from ui

timwessman:
[PR983](https://github.com/City-of-Helsinki/kerrokantasi-ui/pull/983) fixed language change from query params, but it broke language change from UI. So this PR fix also that.


mikkojamG (approved):
Works! 👍

PR #989, opened 2023-11-27
feat: New progress bar to replace old bootstrap one

arpargo:
# Replacing old Bootstrap ProgressBar

### KER-299

----------------------------------------------------------------------------------------------

Reason behind change:
Removing old, deprecated React-bootstrap library from use and switching to HDS components. HDS did not offer suitable component for replacing the ProgressBar so instead created a custom component to solve this need.

### DIFF
--- a/assets/sass/kerrokantasi/_common.scss
+++ b/assets/sass/kerrokantasi/_common.scss
@@ -252,3 +252,17 @@ a.list-group-item {
 .icon-adjustment {
   top: $font-size-base;
 }
+
+.progressBar-wrapper {
+  height: 1.4em;
+  position: relative;
+  width: 100%;
+  border-radius: 5px;
+  background: $light-gray;
+  .progressBar {
+    height: 100%;
+    position: relative;
+    border-radius: inherit;
+    background: blue;
+  }
+}
--- a/assets/sass/kerrokantasi/_common.scss
+++ b/assets/sass/kerrokantasi/_common.scss
@@ -264,5 +264,8 @@ a.list-group-item {
     position: relative;
     border-radius: inherit;
     background: blue;
+    &.in-viewport {
+      transition: width 2s;
+    }
   }
 }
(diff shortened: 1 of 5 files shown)


src/components/QuestionResults.jsx
```diff
@@ -2,7 +2,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { v1 as uuid } from 'uuid';
-import { ProgressBar } from 'react-bootstrap';
+import ProgressBar from './ProgressBar';
```
mikkojamG, line 5: Lint errors here, import order 👀

src/components/admin/QuestionForm.jsx
```diff
@@ -3,7 +3,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { v1 as uuid } from 'uuid';
-import { ProgressBar } from 'react-bootstrap';
+import ProgressBar from '../ProgressBar';
```
mikkojamG, line 6: Lint errors here, import order 👀

assets/sass/kerrokantasi/_common.scss
```diff
@@ -251,4 +251,18 @@ a.list-group-item {
...
+  height: 1.4em;
+  position: relative;
+  width: 100%;
+  border-radius: 5px;
+  background: $light-gray;
+  .progressBar {
+    height: 100%;
+    position: relative;
+    border-radius: inherit;
+    background: blue;
```
mikkojamG, line 266: I think should use --color-coat-of-arms to keep original color 🤔

mikkojamG (changes requested):
Couple of comments, nothing major 🙂

assets/sass/kerrokantasi/_common.scss
arpargo: Actually the original color comes straight from bootstrap but you're right, I should add Helsinki theme styles and use coat-of-arms there 👍

PR #1024, opened 2024-04-26
ci: stop running commitlint against main branch

tuomas777:
Release-please might create commit messages with longer lines than the limit of 72 chars, for example https://github.com/City-of-Helsinki/kerrokantasi-ui/commit/26307af785dd2d5769eac38a05f4bd2019194a9c. That shouldn't be a big issue. However, then [commitlint makes CI action fail](https://github.com/City-of-Helsinki/kerrokantasi-ui/actions/runs/8844826904/job/24287473208), which is not nice.

So disabled running commitlint against the main branch. When a commit is in the main branch it cannot be changed anyway 🙂

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,9 +21,6 @@ jobs:
           cache: 'yarn'
       - run: yarn install --frozen-lockfile
       - run: yarn lint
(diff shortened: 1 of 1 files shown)


arpargo (approved):
LGTM

PR #1274, opened 2026-04-08
WIP PLATTA-7042

tuomas777:


Dockerfile
```diff
@@ -1,104 +1,48 @@
...
-
-USER 1001
-
-CMD ["/bin/bash", "-c", "/usr/share/nginx/html/env.sh && nginx -g \"daemon off;\""]
-
-# Expose port 8086
-EXPOSE 8086
+# - env.sh      (Inherited from base image at /usr/share/nginx/html/env.sh)
+# - USER 1001   (Inherited from base image)
+# - EXPOSE 8080 (Inherited from base image)
```
mikkojamG, line 49: The old port 8086 is still referenced in some places (vite.config.mjs, compose.yml, README.md, playwright.config.js, assets/url.json), should be changed to 8080 🤔 Might bump into different ports in other applications too, like servicemap-ui.

Dockerfile
```diff
@@ -1,104 +1,48 @@
...
+# 1. Install dependencies
+# Base already has /app as WORKDIR
+COPY --chown=default:root package.json yarn.lock ./
+COPY --chown=default:root ./scripts ./scripts
+COPY --chown=default:root ./public ./public
+COPY --chown=default:root ./cities ./cities
+COPY --chown=default:root ./assets ./assets
+
+# 2. Run the install
+RUN yarn --frozen-lockfile --ignore-engines --network-concurrency 1 && yarn cache clean --force
```
mikkojamG: Looks to me --ignore-scripts flag found in old implementation is missing, maybe mixed up with --ignore-engines? 🤔

mikkojamG (changes requested):

Dockerfile
```diff
@@ -1,104 +1,48 @@
...
-COPY --from=staticbuilder /app/nginx.conf /etc/nginx/nginx.conf
-RUN mkdir /etc/nginx/env
-COPY .prod/nginx_env.conf  /etc/nginx/env/
-
+# 2. Setup Runtime Env Injection
+# env.sh is provided by the base image
 WORKDIR /usr/share/nginx/html
-
-# Copy default environment config and setup script
-COPY ./scripts/env.sh .
```
jorilindell, line 91: Maybe the /scripts/env.sh file can be removed since it's no longer used after this change. The reference in README.md should also be removed or updated.

Dockerfile
KannappanSomu: i have replaced all port references to 8080

Dockerfile
KannappanSomu: removed now

Dockerfile
KannappanSomu: Many legitimate packages use scripts to compile native code or perform necessary setup.

The Problem: If we use --ignore-scripts, these packages will be installed but will not work, leading to cryptic "module not found" or "binary missing" errors during yarn build.

mikkojamG: Ignoring scripts also help mitigating supply-chain attacks. Allowing actions eg. postinstall can lead to third-party packages executing malicious code. 🤔

mikkojamG: Ignoring scripts affecting package compiling is also relevant concern, although this hasn't happened so far. In the current situation where supply-chain attacks are very common, I think ignoring scripts by default adds extra layer of protection 🙂

KannappanSomu: ignore scripts flag was ignoring the postinstall script  `yarn update-runtime-env` from executing , so i had to explicitly mention in command like below .. 


```
RUN yarn --frozen-lockfile --ignore-engines --ignore-scripts --network-concurrency 1 \
 && **yarn update-runtime-env** \
 && yarn cache clean --force
```

KannappanSomu: can we close this thread now ?

mikkojamG: Yes. 👍

tuomas777:
Superseded by https://github.com/City-of-Helsinki/kerrokantasi-ui/pull/1280

PR #1291, opened 2026-05-06
fix: main navigation KER-613

Riippi:
In main navigation moving between hearing and map didn't work.

https://helsinkisolutionoffice.atlassian.net/browse/KER-613

Refs: KER-613


mrTuomoK (approved):
ok

PR #1306, opened 2026-05-19
fix: upgrade base images, nodejs and pnpm

tjgofore:
Refs: RATY-319

## Summary by CodeRabbit

* **Chores**
  * Updated Node.js requirement to version 24.15.0 or higher
  * Updated pnpm package manager to version 11.1.3
  * Updated build and deployment base images for improved performance and security
  * Reorganized dependency management configurations


Dockerfile
```diff
@@ -48,7 +48,7 @@ RUN pnpm build
 # ============================================================
 # STAGE 4: Production Runtime
 # ============================================================
-FROM helsinki.azurecr.io/ubi9/nginx-126-spa-standard AS production
+FROM helsinki.azurecr.io/ubi10/nginx-126-spa-standard AS production
```
jorilindell, line 51: Running `docker compose up` on macOS results in `Fatal glibc error: CPU does not support x86-64-v3`. This happens on both Apple Silicon (arm64) and Intel Macs (x86‑64).

jorilindell: This issue was resolved after updating Mac OS

jorilindell (approved):

